PR #3, opened 2018-07-17
kubernetes authenticator client works with conjur v5

jtuttle:
This PR updates the conjur-authn-k8s-client to work with the Conjur v5 enterprise appliance.


main.go
```diff
@@ -33,7 +35,9 @@ func main() {
 		}
 	}
 
+	conjurVersion := os.Getenv("CONJUR_VERSION")
```
dustinmm80: Can you add this to the config options in the readme? I can add it to dockerhub after that.

Also, should this have a default? To "5"?

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+# 0.11.0
+
+- Adds support for Conjur v5.
```
dustinmm80: Can you add something like this to the changelog?

`CONJUR_VERSION`: Conjur version (4 or 5, defaults to 5)

that way we tell users in changelog what they're going to have to update, or about new variables.

dustinmm80: We should probably also mention here and elsewhere that they need to use `'4'` and not `4`. Using the int breaks parsers right?

requests.go
```diff
@@ -20,10 +28,24 @@ func LoginRequest(authnURL string, csrBytes []byte) (*http.Request, error) {
...
+		authenticateURL = fmt.Sprintf("%s/%s/%s/authenticate", authnURL, account, url.QueryEscape(username))
+	}
+
+	var body *bytes.Reader
+
+	if conjurVersion == "5" {
+		body = bytes.NewReader(cert)
+	}
+	
+	//infoLogger.Printf("making authn request to %s", authenticateURL)
```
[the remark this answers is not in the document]
jtuttle: We could leave 'em in yeah. I'll uncomment.

dustinmm80:
Last change, you'll need to update the `VERSION` file, the CI scripting uses this.

dustinmm80:
Looks good. Merging!

PR #19, opened 2019-03-08
Client checks cert expiration before authenticating

### LINKED ISSUE #18: Authenticator can be provided an api key timeout variable
This will allow for easier testing and be able to support flexible rotations.

AC:
- [ ] Authenticator has a way to be provided with an env var that is used for the timout setting on the token

jtuttle:
Also fixes #18


pkg/authenticator/authenticator.go
```diff
@@ -16,11 +16,14 @@ import (
 	"net/url"
 	"os"
 	"strings"
+	"time"
 
 	"github.com/fullsailor/pkcs7"
 )
 
+var infoLogger = log.New(os.Stdout, "INFO: ", log.LUTC|log.Ldate|log.Ltime|log.Lshortfile)
```
jtuttle: You can set the displayed info globally with `SetFlags`. Seems like the kind of thing worth doing for the whole client unless you just have this in for temporary debugging purposes.

```
log.SetFlags(log.LUTC | log.Ldate | log.Ltime | log.Lshortfile)
```

izgeri:
a couple of notes:
- [ ] please bump the version for this change - this project is due for a release anyway
- [ ] please build a local docker image and test using the deploy/demo repos in GKE at least before merging - there's guidelines for using a local authenticator client in the demo repo [here](https://github.com/conjurdemos/kubernetes-conjur-demo#development)

pkg/authenticator/authenticator.go
sgnn7: Logging is now all in one module and we can't use one set of flags since we have both an error and info prefix :(

I did move this stuff into it's own shared module though.

RafiS3:
@sgnn7, @jtuttle I have two questions regarding time difference ([link](https://github.com/cyberark/conjur-authn-k8s-client/pull/19/files#diff-cda0b5e20adeb221b0a91aaf35e592a0R167)):
1. Why aren't all the variables that deal with time, being converted to UTC? What happens if `NotAfter` is in CET  and `currentDate ` is in EST?
2. What if Conjur and the authn client do have the same time zone, but don't have a synced clock? I mean that for some reason, one is five minutes ahead of the other?

sgnn7:
@RafiS3 
> Why aren't all the variables that deal with time, being converted to UTC? What happens if NotAfter is in CET and currentDate is in EST?

Made the change to convert them to UTC before comparison. I think the code would have worked as expected even with mismatched time zones but the new change does make this clear.

> What if Conjur and the authn client do have the same time zone, but don't have a synced clock? I mean that for some reason, one is five minutes ahead of the other?
NTP is _very_ important for cluster work and given that the follower that the certs are issued from must run in the same cluster as the app, I don't see this as a common problem. However, if time drift happens to be the case somehow we have two clock drift cases:
- Client clock is faster than the server: Client will do a login earlier than anticipated and unless the clock drift is greater than 3 days, the authenticator will work properly
- Client clock is slower than the server:
  - We have a 30 second buffer time included so 30s drift should  work seamlessly (we can increase this value too). 
  - If we are within the timeout maximum of [2min](https://github.com/cyberark/conjur-authn-k8s-client/blob/master/cmd/authenticator/main.go#L85) + the mentioned 30 seconds, the server will return 401s for a short period of time and then authenticate successfully. 
  - If we are beyond 2.5 minutes slower than server, the authenticator container will fail and the pod will be rescheduled (depending on the deployment too)

pkg/authenticator/authenticator.go
jtuttle: The prefix is the 2nd argument though right? So that can still differ regardless. If we want the same set of flags for info and error then you could set them globally, but maybe we don't.

RafiS3:
Thanks for the response @sgnn7. Regarding:

> If we are beyond 2.5 minutes slower than server, the authenticator container will fail and the pod will be rescheduled (depending on the deployment too)

Until now, if the authenticator container failed, it didn't stop the sidecar container and therefore, didn't cause a pod recycle. Did you change the behaviour part of this PR?

README.md
```diff
@@ -25,14 +25,15 @@ The client is configured entirely through environment variables. These are liste
 - `CONJUR_AUTHN_URL`: URL pointing to authenticator service endpoint
 - `CONJUR_AUTHN_LOGIN`: Host login for pod e.g. `namespace/service_account/some_service_account`
 - `CONJUR_SSL_CERTIFICATE`: Public SSL cert for Conjur connection
+- `CONJUR_TOKEN_TIMEOUT`: Timeout for fetching a new token (defaults to 6 minutes). In most cases, this variable shoudl not be modified.
```
izgeri: shoudl => should

pkg/authenticator/authenticator.go
sgnn7: Resolved in DMs

README.md
RafiS3: I think it's also worth mentioning when and why this parameter should be used. Just to be as clear as possible to the end user.

izgeri:
let's also verify we have actually resolved the customer reported issue by running some tests where:
- [ ] the appliance is hardcoded to refresh the cert after 5 minutes, and we see it go through several cycles and refresh the cert each time (and continue to be able to authenticate)
- [ ] the appliance is hardcoded to refresh the cert after 1 hour, and we see it successfully refresh the cert after 1 hour (and continue to be able to authenticate)
- [TBD] the appliance is umodified, and the cert is automatically refreshed after the 3-day expiration period (and the app is still able to authenticate)

PR #26, opened 2019-08-14
Retrieve k8s Secrets and update them with Conjur secrets

orenbm:
This commit holds the main logic of handling k8s secrets. The flow is:

1. Retrieve all the required secrets from k8s
2. Parse each secret's "conjur-map" entry
3. Retrieve required Conjur secrets and update them in the map
4. Patch the k8s secrets with the values from Conjur

This commit includes also:
- Update Secrets config to read the K8S_SECRETS env var that list the required secrets for the pod
- Remove logic & files of storing secrets in the volume


pkg/secrets/conjur_secrets.go
```diff
@@ -0,0 +1,38 @@
...
+import (
+	"fmt"
+)
+
+type ConjurProvider interface {
+	RetrieveSecret(string) ([]byte, error)
+	RetrieveBatchSecrets([]string) (map[string][]byte, error)
+}
+
+func GetVariableIDsToRetrieve(pathMap map[string]string) []string {
```
nessiLahav: Add a comment what this map contains

pkg/secrets/kubernetes_secrets.go
```diff
@@ -2,55 +2,121 @@ package secrets
...
+	stringDataEntry = nil
 	if err != nil {
-		return fmt.Errorf("Failed to write Kubernetes secret: %s", err)
+		return fmt.Errorf("failed to patch Kubernetes secret: %s", err)
 	}
 
 	return nil
 }
+
+func generateStringDataEntry(stringDataEntriesMap map[string][]byte) []byte {
```
nessiLahav: Add comment about this function purpose

pkg/secrets/secrets.go
```diff
@@ -1,183 +1,129 @@
...
+			return nil, fmt.Errorf("error reading k8s secrets: %s", err)
+		}
 
-func (secrets *Secrets) HandleSecretsResponse(response *SecretResponse) error {
-	InfoLogger.Printf("Writing secrets to destinations...")
+		// Parse its "conjur-map" data entry and store its values in the new-data-entries map
+		// This map holds data entries that will be added to the k8s secret after we retrieve their values from Conjur
+		newDataEntriesMap := make(map[string][]byte)
+		for key, value := range k8sSecret.Secret.Data {
+			if key == "conjur-map" {
```
nessiLahav: "conjur-map" declare it as const in config.go

pkg/secrets/kubernetes_secrets.go
```diff
@@ -2,55 +2,130 @@ package secrets
...
-	// Create the Kubernetes secret
-	secret := v1.Secret{
-		ObjectMeta: metav1.ObjectMeta{
-			Name: secretName,
-		},
-		Data: secretsData,
+	// TODO: extract until here
+
+	InfoLogger.Printf("Retrieving Kubernetes secret '%s' from namespace '%s'...", secretName, namespace)
+	k8sSecret, err := kubeClient.CoreV1().Secrets(namespace).Get(secretName, metav1.GetOptions{})
```
sigalsax: Here we are assuming that all of customer's Conjur secrets will be wrapped up as data entries in a single K8s secret?

orenbm: we don't. but each k8s secret (that needs secrets from conjur) will have a "conjur-map" entry which will hold the required secrets *for that k8s secret*. we may have more than one k8s secret so we run the operation above for each one

pkg/secrets/kubernetes_secrets.go
sigalsax: desc here is very helpful

PR #30, opened 2019-08-21
WIPb: Add initial base for k8s secrets

sigalsax:
This PR adds the initial base (creating structs, interfaces, and initial flow implementation) for later development of design for the K8s secrets, allowing the later implementation process to be divided amongst engineers.

### DIFF
--- a/cmd/sidecar/main.go
+++ b/cmd/sidecar/main.go
@@ -9,6 +9,8 @@ import (
 	"github.com/cenkalti/backoff"
 	"github.com/cyberark/conjur-authn-k8s-client/pkg/authenticator"
 	authnConfig "github.com/cyberark/conjur-authn-k8s-client/pkg/authenticator/config"
+	"github.com/cyberark/conjur-authn-k8s-client/pkg/storage"
+	storageConfig "github.com/cyberark/conjur-authn-k8s-client/pkg/storage/config"
 	"github.com/cyberark/conjur-authn-k8s-client/pkg/secrets"
 	secretsConfig "github.com/cyberark/conjur-authn-k8s-client/pkg/secrets/config"
 	log "github.com/cyberark/conjur-authn-k8s-client/pkg/sidecar/logging"
@@ -20,6 +22,10 @@ var infoLogger = log.InfoLogger
 
 func main() {
 	var err error
+	var secretsHandler *secrets.Secrets
+
+	//SECRETS_DESTINATION read from env variables
+	conjurSecretsDest := 0
 
 	// Parse any flags for client cert / token paths, and set default values if not passed
 	clientCertPath := flag.String("c", "/etc/conjur/ssl/client.pem",
@@ -40,17 +46,33 @@ func main() {
 		os.Exit(1)
 	}
 
-	configSecrets, err := secretsConfig.NewFromEnv(tokenFilePath)
+	configStorage, err := storageConfig.NewFromEnv()
 	if err != nil {
 		errLogger.Printf(err.Error())
 		os.Exit(1)
 	}
+	// Create new Storage
+	stor, err := storage.NewStorageHandler(*configStorage)
 
-	// Create new Secrets
-	secrets, err := secrets.New(*configSecrets)
-	if err != nil {
-		errLogger.Printf(err.Error())
-		os.Exit(1)
+	//if conjurSecretsDest == 1  &&  authn.Config.ContainerMode != "init" {
+	//      // notify on invalid configuration and exit
+	//      errLogger.Printf(" appropriate message for not supporting sidecar ....")
+	//      os.Exit(1)
+	//}
+
+	if conjurSecretsDest == 1 {
+		configSecrets, err := secretsConfig.NewFromEnv(tokenFilePath)
+		if err != nil {
+			errLogger.Printf(err.Error())
+			os.Exit(1)
+		}
+
+		// Create new Secrets
+		secretsHandler, err = secrets.New(*configSecrets)
+		if err != nil {
+			errLogger.Printf(err.Error())
+			os.Exit(1)
+		}
 	}
 
 	// Configure exponential backoff
@@ -76,27 +98,28 @@ func main() {
 				return err
 			}
 
-			// ------------ START LOGIC ------------
+			if conjurSecretsDest == 1 {
 
-			k8sSecretsMap, err := secrets.RetrieveK8sSecrets()
-			if err != nil {
-				errLogger.Printf("Failed to retrieve k8s secrets: %s", err.Error())
-				return err
+				// ------------ START LOGIC ------------
+				k8sSecretsMap, err := secretsHandler.RetrieveK8sSecrets()
+				if err != nil {
+					errLogger.Printf("Failed to retrieve k8s secrets: %s", err.Error())
+					return err
+				}
+
+				k8sSecretsMap, err = secretsHandler.UpdateK8sSecretsMapWithConjurSecrets(k8sSecretsMap)
+				if err != nil {
+					errLogger.Printf("Failed to update K8s Secrets map: %s", err.Error())
+					return err
+				}
+
+				err = secretsHandler.PatchK8sSecrets(k8sSecretsMap)
+				if err != nil {
+					errLogger.Printf("Failed to patch K8s Secrets: %s", err.Error())
+					return err
+				}
+				// ------------ END LOGIC ------------
 			}
-
-			k8sSecretsMap, err = secrets.UpdateK8sSecretsMapWithConjurSecrets(k8sSecretsMap)
-			if err != nil {
-				errLogger.Printf("Failed to update K8s Secrets map: %s", err.Error())
-				return err
-			}
-
-			err = secrets.PatchK8sSecrets(k8sSecretsMap)
-			if err != nil {
-				errLogger.Printf("Failed to patch K8s Secrets: %s", err.Error())
-				return err
-			}
-
-			// ------------ END LOGIC ------------
 
 			if authn.Config.ContainerMode == "init" {
 				os.Exit(0)
--- a/cmd/sidecar/main.go
+++ b/cmd/sidecar/main.go
@@ -9,7 +9,7 @@ import (
 	"github.com/cenkalti/backoff"
 	"github.com/cyberark/conjur-authn-k8s-client/pkg/authenticator"
 	authnConfig "github.com/cyberark/conjur-authn-k8s-client/pkg/authenticator/config"
-	storageHandler "github.com/cyberark/conjur-authn-k8s-client/pkg/storage"
+	storage "github.com/cyberark/conjur-authn-k8s-client/pkg/storage"
 	storageConfig "github.com/cyberark/conjur-authn-k8s-client/pkg/storage/config"
 	"github.com/cyberark/conjur-authn-k8s-client/pkg/secrets"
 	secretsConfig "github.com/cyberark/conjur-authn-k8s-client/pkg/secrets/config"
@@ -24,11 +24,6 @@ func main() {
 	var err error
 	var secretsHandler *secrets.Secrets
 
-	const(
-		none = "none"
-		k8s = "k8s_secrets"
-	)
-
 	// Parse any flags for client cert / token paths, and set default values if not passed
 	clientCertPath := flag.String("c", "/etc/conjur/ssl/client.pem",
 		"Path to client certificate")
@@ -39,10 +34,6 @@ func main() {
 	if err != nil {
 		errLogger.Printf(err.Error())
 		os.Exit(1)
-	}
-
-	if configAuthn.ContainerMode == "init" {
-		os.Exit(0)
 	}
 
 	// Create new Authenticator
@@ -58,14 +49,14 @@ func main() {
 		errLogger.Printf(err.Error())
 		os.Exit(1)
 	}
-	stor, err := storageHandler.NewStorageHandler(*configStorage)
+	storageHandler, err := storage.NewStorageHandler(*configStorage)
 	if err != nil {
 		errLogger.Printf(err.Error())
 		os.Exit(1)
 	}
 
 	// TODO: account for if SECRET_DESTINATION is set to k8s but not an init container. Must check this case early.
-	conjurSecretsDest := none
+	conjurSecretsDest := storageConfig.K8s
 	/* if conjurSecretsDest == 1  &&  authn.Config.ContainerMode != "init" {
 	     // notify on invalid configuration and exit
 	     infoLogger.Printf("Appropriate message for not supporting sidecar with this workflow ....")
@@ -74,7 +65,7 @@ func main() {
 
 	// TODO: move these code segments to their respective files
 	//  Code that will handle the placement of access token
-	if conjurSecretsDest == none {
+	if conjurSecretsDest == storageConfig.None {
 		configSecrets, err := secretsConfig.NewFromEnv(tokenFilePath)
 		if err != nil {
 			errLogger.Printf("Failure creating secrets config: %s", err.Error())
@@ -133,6 +124,10 @@ func main() {
 				return err
 			}
 			// ------------ END LOGIC ------------
+
+			if configAuthn.ContainerMode == "init" {
+				os.Exit(0)
+			}
 
 			// Reset exponential backoff
 			expBackoff.Reset()
--- a/cmd/sidecar/main.go
+++ b/cmd/sidecar/main.go
@@ -9,7 +9,7 @@ import (
 	"github.com/cenkalti/backoff"
 	"github.com/cyberark/conjur-authn-k8s-client/pkg/authenticator"
 	authnConfig "github.com/cyberark/conjur-authn-k8s-client/pkg/authenticator/config"
-	"github.com/cyberark/conjur-authn-k8s-client/pkg/storage"
+	storageHandler "github.com/cyberark/conjur-authn-k8s-client/pkg/storage"
 	storageConfig "github.com/cyberark/conjur-authn-k8s-client/pkg/storage/config"
 	"github.com/cyberark/conjur-authn-k8s-client/pkg/secrets"
 	secretsConfig "github.com/cyberark/conjur-authn-k8s-client/pkg/secrets/config"
@@ -24,8 +24,10 @@ func main() {
 	var err error
 	var secretsHandler *secrets.Secrets
 
-	//SECRETS_DESTINATION read from env variables
-	conjurSecretsDest := 0
+	const(
+		none = "none"
+		k8s = "k8s_secrets"
+	)
 
 	// Parse any flags for client cert / token paths, and set default values if not passed
 	clientCertPath := flag.String("c", "/etc/conjur/ssl/client.pem",
@@ -39,6 +41,10 @@ func main() {
 		os.Exit(1)
 	}
 
+	if configAuthn.ContainerMode == "init" {
+		os.Exit(0)
+	}
+
 	// Create new Authenticator
 	authn, err := authenticator.New(*configAuthn)
 	if err != nil {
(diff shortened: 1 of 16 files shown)


cmd/sidecar/main.go
```diff
@@ -20,6 +22,10 @@ var infoLogger = log.InfoLogger
 
 func main() {
 	var err error
+	var secretsHandler *secrets.Secrets
+
+	//SECRETS_DESTINATION read from env variables
```
nessiLahav: Add "TODO" comment to be clear how it should be implemented later

cmd/sidecar/main.go
```diff
@@ -40,17 +46,33 @@ func main() {
...
+	// Create new Storage
+	stor, err := storage.NewStorageHandler(*configStorage)
+
+	//if conjurSecretsDest == 1  &&  authn.Config.ContainerMode != "init" {
+	//      // notify on invalid configuration and exit
+	//      errLogger.Printf(" appropriate message for not supporting sidecar ....")
+	//      os.Exit(1)
+	//}
+
+	if conjurSecretsDest == 1 {
```
orenbm: what is `1`? can we have some const for this?

cmd/sidecar/main.go
```diff
@@ -40,17 +46,33 @@ func main() {
...
-	configSecrets, err := secretsConfig.NewFromEnv(tokenFilePath)
+	configStorage, err := storageConfig.NewFromEnv()
 	if err != nil {
 		errLogger.Printf(err.Error())
 		os.Exit(1)
 	}
+	// Create new Storage
+	stor, err := storage.NewStorageHandler(*configStorage)
+
+	//if conjurSecretsDest == 1  &&  authn.Config.ContainerMode != "init" {
```
nessiLahav: Also here add "TODO"

cmd/sidecar/main.go
```diff
@@ -40,17 +46,33 @@ func main() {
...
+	//      // notify on invalid configuration and exit
+	//      errLogger.Printf(" appropriate message for not supporting sidecar ....")
+	//      os.Exit(1)
+	//}
+
+	if conjurSecretsDest == 1 {
+		configSecrets, err := secretsConfig.NewFromEnv(tokenFilePath)
+		if err != nil {
+			errLogger.Printf(err.Error())
```
orenbm: please some text here so we can pinpoint it easier in the future. for example:
`errLogger.Printf("failed to create secrets config: %s", err.Error())`

cmd/sidecar/main.go
```diff
@@ -40,17 +46,33 @@ func main() {
...
 	}
 
-	configSecrets, err := secretsConfig.NewFromEnv(tokenFilePath)
+	configStorage, err := storageConfig.NewFromEnv()
 	if err != nil {
 		errLogger.Printf(err.Error())
 		os.Exit(1)
 	}
+	// Create new Storage
+	stor, err := storage.NewStorageHandler(*configStorage)
```
orenbm: please refactor to `storageHandler`

cmd/sidecar/main.go
```diff
@@ -40,17 +46,33 @@ func main() {
...
 
-	// Create new Secrets
-	secrets, err := secrets.New(*configSecrets)
-	if err != nil {
-		errLogger.Printf(err.Error())
-		os.Exit(1)
+		// Create new Secrets
+		secretsHandler, err = secrets.New(*configSecrets)
+		if err != nil {
+			errLogger.Printf(err.Error())
```
orenbm: same here

pkg/storage/access_token_file.go
```diff
@@ -0,0 +1,25 @@
...
+type AccessTokenFile struct {
+	TokenFilePath	string
+	Data	[]byte
+}
+
+func NewAccessTokenFile(config config.Config) (token *AccessTokenFile, err error) {
+	return
+}
+
+func (token *AccessTokenFile) Read() (Data []byte, err error) {
```
nessiLahav: If we are not changing the struct it should pass by value not by reference (without *)

cmd/sidecar/main.go
```diff
@@ -76,28 +98,29 @@ func main() {
...
 
-			err = secrets.PatchK8sSecrets(k8sSecretsMap)
-			if err != nil {
-				errLogger.Printf("Failed to patch K8s Secrets: %s", err.Error())
-				return err
-			}
-
-			// ------------ END LOGIC ------------
-
 			if authn.Config.ContainerMode == "init" {
```
orenbm: i think this should move upwards so we fail early on this

pkg/storage/access_token_file.go
nessiLahav: also here pass by value

pkg/storage/access_token_memory.go
```diff
@@ -0,0 +1,33 @@
...
+
+// New returns a new AccessTokenMemory object
+// return interface
+func NewAccessTokenMemory(config config.Config) (token *AccessTokenMemory, err error) {
+	return &AccessTokenMemory {
+
+	}, nil
+}
+
+func (token *AccessTokenMemory) Read() (Data []byte, err error) {
```
nessiLahav: also here pass by value

pkg/storage/secrets_fetcher.go
```diff
@@ -0,0 +1,3 @@
+package storage
+
+//logic of fetch sercets from Conjur
```
orenbm: please rename to `conjur_secrets_fetcher`

orenbm: and also move it to `secrets` package

pkg/storage/secrets_handler_k8s.go
```diff
@@ -0,0 +1,25 @@
+package storage
+
+import (
+	"github.com/cyberark/conjur-authn-k8s-client/pkg/storage/config"
+)
+
+type SecretsFetcher interface {
```
orenbm: ConjurSecretsFetcher

pkg/storage/storage_handler.go
```diff
@@ -0,0 +1,44 @@
...
+	HandleSecrets() error
+}
+
+type AccessTokenHandler interface {
+	Read()	([]byte, error)
+	Write()	error
+	Delete()	error
+}
+
+// Need underscore here unsure why
```
orenbm: please add a todo to understand this

pkg/storage/storage_handler.go
```diff
@@ -0,0 +1,44 @@
...
+type Storage_Handler struct {
+	AccessToken AccessTokenHandler
+	Secrets SecretsHandler
+}
+
+func NewStorageHandler(config config.Config) (StorageHandler *Storage_Handler, err error) {
+	// check what StoreType was in original store
+	if config.StoreType == "k8s_secrets" {
```
orenbm: don't we have a const for this?

cmd/sidecar/main.go
moticless: ... suppose to be something like:
	conjurSecretsDest := os.Getenv("CONJUR_SECRETS_DESTINATION")
	if len(conjurSecretsDest) == 0 {
		conjurSecretsDest = "none"
 	} else {
                  // TODO: Validate value is either "none" or "k8s_secrets". Otherwise output an error
	}
	errLogger.Printf("CONJUR_SECRETS_DESTINATION: %s", conjurSecretsDest)

pkg/storage/storage_handler.go
```diff
@@ -0,0 +1,44 @@
...
+		// create new AccessTokenMemory object
+		tokenInMemory, err := NewAccessTokenMemory(config)
+		// create new SecretHandler object
+		secretsHandlerK8s, err := NewSecretHandlerK8s(config, tokenInMemory)
```
orenbm: you take the error without checking it

pkg/storage/storage_handler.go
```diff
@@ -0,0 +1,44 @@
...
+		return &Storage_Handler{
+			AccessToken: tokenInMemory,
+			Secrets:     secretsHandlerK8s,
+		}, nil
+	} else {
+		//	then the StoreType is none and old way of saving access token in file is run
+		// check if initContainer or sidecar here or at the beginning?
+		tokenInFile, err := NewAccessTokenFile(config)
+		secretsHandlerNone, err := NewSecretHandlerNone(config, tokenInFile)
```
orenbm: you take the error without checking it

pkg/storage/config/config.go
```diff
@@ -0,0 +1,30 @@
...
+const (
+	k8s  = "k8s_secrets"
+	none = "none"
+)
+
+// should we check here what the ENV says and act accordingly with values?
+func NewFromEnv() (*Config, error) {
+	storeType := none
+	tokenFilePath := "run/conjur/access-token"
+	if os.Getenv("SECRETS_DESTINATION") == k8s {
```
nessiLahav: Add "TODO" comment how it should be implemented according to the design: 
k8s - new logic
none or not exists - old logic
incorrect - error

cmd/sidecar/main.go
```diff
@@ -40,17 +46,33 @@ func main() {
...
+	stor, err := storage.NewStorageHandler(*configStorage)
+
+	//if conjurSecretsDest == 1  &&  authn.Config.ContainerMode != "init" {
+	//      // notify on invalid configuration and exit
+	//      errLogger.Printf(" appropriate message for not supporting sidecar ....")
+	//      os.Exit(1)
+	//}
+
+	if conjurSecretsDest == 1 {
+		configSecrets, err := secretsConfig.NewFromEnv(tokenFilePath)
```
sigalsax: @orenbm @nessiLahav @moticless I know we are slowing moving this logic over to different files but is this logic correct here? if i am understanding correctly, if `secrets_destination===1` (k8s) then read from token in set path. This is not what we want here b/c we want to read the access token from memory. correct?

cmd/sidecar/main.go
```diff
@@ -76,28 +98,29 @@ func main() {
...
-			if err != nil {
-				errLogger.Printf("Failed to retrieve k8s secrets: %s", err.Error())
-				return err
-			}
-
-			k8sSecretsMap, err = secrets.UpdateK8sSecretsMapWithConjurSecrets(k8sSecretsMap)
-			if err != nil {
-				errLogger.Printf("Failed to update K8s Secrets map: %s", err.Error())
-				return err
+			if conjurSecretsDest == 1 {
```
orenbm: i don't think we need this `if` clause. we already initialized `secretsHandler` according to the value of `conjurSecretsDest`. now we only need to run `secretsHandler.HandleSecrets`. am i wrong?

cmd/sidecar/main.go
orenbm: also it seems that we can check this on `configAuthn` instead of `authn.config`

pkg/storage/access_token_memory.go
```diff
@@ -0,0 +1,33 @@
+package storage
+
+import "github.com/cyberark/conjur-authn-k8s-client/pkg/storage/config"
+
+//type AccessTokenHandler interface {
```
orenbm: what's this? please remove if redundant of comment a TODO if not

pkg/storage/access_token_memory.go
```diff
@@ -0,0 +1,33 @@
...
+//	Write()	error
+//	Delete()	error
+//}
+
+type AccessTokenMemory struct {
+	Data	[]byte
+}
+
+// New returns a new AccessTokenMemory object
+// return interface
```
orenbm: what? does it return an object or an interface?

cmd/sidecar/main.go
sigalsax: @orenbm @nessiLahav @moticless I know we are slowing moving this logic over to different files but is this logic correct here? if i am understanding correctly, if secrets_destination===1 (k8s) then read from token in set path. This is not what we want here b/c we want to read the access token from memory. correct?

cmd/sidecar/main.go
orenbm: the last line should be logged with `infoLogger`, not `errLogger`

orenbm: also, loading `os.Getenv("CONJUR_SECRETS_DESTINATION")` should be done in the config file, not in main

pkg/storage/storage_handler.go
```diff
@@ -0,0 +1,44 @@
...
+
+type AccessTokenHandler interface {
+	Read()	([]byte, error)
+	Write()	error
+	Delete()	error
+}
+
+// Need underscore here unsure why
+type Storage_Handler struct {
+	AccessToken AccessTokenHandler
```
nessiLahav: Defined the members as pointers like you did in all other structs

pkg/storage/config/config.go
orenbm: please let's not call it "old logic" and "new logic"

cmd/sidecar/main.go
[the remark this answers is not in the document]
sigalsax: should have been a description but realize it doesnt contribute to anything. removing

sigalsax: removing because it doesn't contribute to anything

cmd/sidecar/main.go
orenbm: in general, yes. maybe it should be a switch case but the logic itself is correct

cmd/sidecar/main.go
sigalsax: updated but kept the logic flow with a detailed header because this logic will be moved to their respective files and most probably will be changed again

cmd/sidecar/main.go
```diff
@@ -40,17 +46,33 @@ func main() {
 		os.Exit(1)
 	}
 
-	configSecrets, err := secretsConfig.NewFromEnv(tokenFilePath)
```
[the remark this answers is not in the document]
sigalsax: say again?

cmd/sidecar/main.go
sigalsax: just spoke about this with you and we decided to add a TODO at the top

pkg/storage/storage_handler.go
sigalsax: After discussion with Moti we realize that interfaces do not need to be specified with a * because they are implicit

pkg/storage/config/config.go
```diff
@@ -0,0 +1,30 @@
...
+	TokenFilePath string
+}
+
+const (
+	k8s  = "k8s_secrets"
+	none = "none"
+)
+
+// should we check here what the ENV says and act accordingly with values?
+func NewFromEnv() (*Config, error) {
```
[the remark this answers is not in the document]
sigalsax: because we want to keep it reuseable. This NewFromEnv will see both flows- for k8s and none. For k8s, the initialization would `storageConfig.NewFromEnv()` and not storageConfig.NewFromEnv(""). We could definitely demand that a param is passed in (`""` or `"/run/conjur/access-token"`). Let me know what you think.

pkg/storage/config/config.go
```diff
@@ -0,0 +1,30 @@
...
+	storeType := none
+	tokenFilePath := "run/conjur/access-token"
+	if os.Getenv("SECRETS_DESTINATION") == k8s {
+		storeType = k8s
+		tokenFilePath = ""
+	}
+	// else old flow
```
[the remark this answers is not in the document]
sigalsax: not a todo, but explaining what is happening in the next code block. I removed this because it was meant to be helpful for me to follow the code

cmd/sidecar/main.go
```diff
@@ -40,17 +46,33 @@ func main() {
...
 	}
+	// Create new Storage
+	stor, err := storage.NewStorageHandler(*configStorage)
+
+	//if conjurSecretsDest == 1  &&  authn.Config.ContainerMode != "init" {
+	//      // notify on invalid configuration and exit
+	//      errLogger.Printf(" appropriate message for not supporting sidecar ....")
+	//      os.Exit(1)
+	//}
+
```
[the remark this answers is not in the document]
sigalsax: added explanation that need to check scenario where `SECRET_DESTINATION==k8s` but not a init container. This is something we don't support

cmd/sidecar/main.go
sigalsax: added const to top of file. Also added above the associated code both that this is a TODO

cmd/sidecar/main.go
sigalsax: added but I do think we should keep it as general as possible because this could change when we actually move these code blocks to their files

pkg/storage/storage_handler.go
sigalsax: moved this below where it was initalized but what are we checking here? 

_Big picture question:_ Why do we care if authenticator is done as an init?

pkg/storage/secrets_handler_k8s.go
sigalsax: renamed and updated naming across files

pkg/storage/storage_handler.go
sigalsax: good point. added!

pkg/storage/storage_handler.go
sigalsax: in a nested package. Do we want to add another const here? or is there a way to use the same const across different packages?

cmd/sidecar/main.go
sigalsax: Doesnt the k8s flow read the access token from memory not from a file? The above example states that if k8s flow, read from token in set path (which I assume to be a file at that defined location)

cmd/sidecar/main.go
```diff
@@ -33,26 +41,54 @@ func main() {
...
+		configSecrets, err := secretsConfig.NewFromEnv(tokenFilePath)
+		if err != nil {
+			errLogger.Printf("Failure creating secrets config: %s", err.Error())
+			os.Exit(1)
+		}
+
+		// Create new Secrets
+		secretsHandler, err = secrets.New(*configSecrets)
+		if err != nil {
+			errLogger.Printf("Failure creating secrets: %s", err.Error())
```
sigalsax: do I have to put a this error code block whenever we expect an error to be returned? @orenbm

cmd/sidecar/main.go
```diff
@@ -20,6 +22,12 @@ var infoLogger = log.InfoLogger
 
 func main() {
 	var err error
+	var secretsHandler *secrets.Secrets
+
+	const(
```
sigalsax: is there a way to open this const up to other packages? so we can use a single const across our project?

cmd/sidecar/main.go
```diff
@@ -39,38 +41,50 @@ func main() {
...
 	}
-	// Create new Storage
-	stor, err := storage.NewStorageHandler(*configStorage)
-
-	//if conjurSecretsDest == 1  &&  authn.Config.ContainerMode != "init" {
-	//      // notify on invalid configuration and exit
-	//      errLogger.Printf(" appropriate message for not supporting sidecar ....")
-	//      os.Exit(1)
-	//}
+	stor, err := storageHandler.NewStorageHandler(*configStorage)
```
orenbm: change to `storageHandler`

cmd/sidecar/main.go
```diff
@@ -9,7 +9,7 @@ import (
 	"github.com/cenkalti/backoff"
 	"github.com/cyberark/conjur-authn-k8s-client/pkg/authenticator"
 	authnConfig "github.com/cyberark/conjur-authn-k8s-client/pkg/authenticator/config"
-	"github.com/cyberark/conjur-authn-k8s-client/pkg/storage"
+	storageHandler "github.com/cyberark/conjur-authn-k8s-client/pkg/storage"
```
orenbm: i think this should be `storage` (so you can just remove the `storageHandler`). this object creates a hadnler

cmd/sidecar/main.go
orenbm: as long as we think it can assist us

cmd/sidecar/main.go
```diff
@@ -39,38 +41,50 @@ func main() {
 		os.Exit(1)
 	}
 
+	if configAuthn.ContainerMode == "init" {
```
orenbm: this should be right at the end

cmd/sidecar/main.go
sigalsax: *interesting find*, it is not enough to just import, all variables that we want to export from another file need to be in *uppercase*
Ex:
in package `*storageConfig* ...storage/config` we have consts,
`
const (
	K8s  = "k8s_secrets"
	None = "none"
)
`
In order to use each consts outside this pkg , we need to capitalize them

Then in `sidecar/main.go` we import `*storageConfig* ...storage/config` and write storageConfig.K8s.

v cool stuff

pkg/storage/storage_handler.go
sigalsax: did et

PR #31, opened 2019-08-22
Impl access token

nessiLahav:
#### What does this PR do?
implement AccessToken interface and add UT

#### How to run?
run the following command `./bin/test`

### DIFF
--- a/cmd/sidecar/main.go
+++ b/cmd/sidecar/main.go
@@ -9,11 +9,11 @@ import (
 	"github.com/cenkalti/backoff"
 	"github.com/cyberark/conjur-authn-k8s-client/pkg/authenticator"
 	authnConfig "github.com/cyberark/conjur-authn-k8s-client/pkg/authenticator/config"
-	storage "github.com/cyberark/conjur-authn-k8s-client/pkg/storage"
-	storageConfig "github.com/cyberark/conjur-authn-k8s-client/pkg/storage/config"
 	"github.com/cyberark/conjur-authn-k8s-client/pkg/secrets"
 	secretsConfig "github.com/cyberark/conjur-authn-k8s-client/pkg/secrets/config"
 	log "github.com/cyberark/conjur-authn-k8s-client/pkg/sidecar/logging"
+	storage "github.com/cyberark/conjur-authn-k8s-client/pkg/storage"
+	storageConfig "github.com/cyberark/conjur-authn-k8s-client/pkg/storage/config"
 )
 
 // logging
@@ -30,33 +30,34 @@ func main() {
 	tokenFilePath := flag.String("t", "/run/conjur/access-token",
 		"Path to Conjur access token")
 
-	configAuthn, err := authnConfig.NewFromEnv(clientCertPath, tokenFilePath)
-	if err != nil {
-		errLogger.Printf(err.Error())
-		os.Exit(1)
-	}
-
-	// Create new Authenticator
-	authn, err := authenticator.New(*configAuthn)
-	if err != nil {
-		errLogger.Printf(err.Error())
-		os.Exit(1)
-	}
-
 	// Create new Storage
-	configStorage, err := storageConfig.NewFromEnv()
+	configStorage, err := storageConfig.NewFromEnv(tokenFilePath)
 	if err != nil {
 		errLogger.Printf(err.Error())
 		os.Exit(1)
 	}
+
 	storageHandler, err := storage.NewStorageHandler(*configStorage)
 	if err != nil {
 		errLogger.Printf(err.Error())
 		os.Exit(1)
 	}
 
+	configAuthn, err := authnConfig.NewFromEnv(clientCertPath)
+	if err != nil {
+		errLogger.Printf(err.Error())
+		os.Exit(1)
+	}
+
+	// Create new Authenticator
+	authn, err := authenticator.New(*configAuthn, storageHandler.AccessToken)
+	if err != nil {
+		errLogger.Printf(err.Error())
+		os.Exit(1)
+	}
+
 	// TODO: account for if SECRET_DESTINATION is set to k8s but not an init container. Must check this case early.
-	conjurSecretsDest := storageConfig.K8s
+	conjurSecretsDest := storageConfig.K8S
 	/* if conjurSecretsDest == 1  &&  authn.Config.ContainerMode != "init" {
 	     // notify on invalid configuration and exit
 	     infoLogger.Printf("Appropriate message for not supporting sidecar with this workflow ....")
--- a/cmd/sidecar/main.go
+++ b/cmd/sidecar/main.go
@@ -1,7 +1,6 @@
 package main
 
 import (
-	"flag"
 	"fmt"
 	"os"
 	"time"
@@ -12,7 +11,7 @@ import (
 	"github.com/cyberark/conjur-authn-k8s-client/pkg/secrets"
 	secretsConfig "github.com/cyberark/conjur-authn-k8s-client/pkg/secrets/config"
 	log "github.com/cyberark/conjur-authn-k8s-client/pkg/sidecar/logging"
-	storage "github.com/cyberark/conjur-authn-k8s-client/pkg/storage"
+	"github.com/cyberark/conjur-authn-k8s-client/pkg/storage"
 	storageConfig "github.com/cyberark/conjur-authn-k8s-client/pkg/storage/config"
 )
 
@@ -24,14 +23,8 @@ func main() {
 	var err error
 	var secretsHandler *secrets.Secrets
 
-	// Parse any flags for client cert / token paths, and set default values if not passed
-	clientCertPath := flag.String("c", "/etc/conjur/ssl/client.pem",
-		"Path to client certificate")
-	tokenFilePath := flag.String("t", "/run/conjur/access-token",
-		"Path to Conjur access token")
-
 	// Create new Storage
-	configStorage, err := storageConfig.NewFromEnv(tokenFilePath)
+	configStorage, err := storageConfig.NewFromEnv()
 	if err != nil {
 		errLogger.Printf(err.Error())
 		os.Exit(1)
@@ -43,7 +36,7 @@ func main() {
 		os.Exit(1)
 	}
 
-	configAuthn, err := authnConfig.NewFromEnv(clientCertPath)
+	configAuthn, err := authnConfig.NewFromEnv()
 	if err != nil {
 		errLogger.Printf(err.Error())
 		os.Exit(1)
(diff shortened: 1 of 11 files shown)


pkg/storage/access_token_memory.go
```diff
@@ -1,25 +1,42 @@
...
+	}
+
+	token.Data = Data
 	return nil
 }
 
 func (token *AccessTokenMemory) Delete() (err error) {
+	// Clear Data
+	empty := make([]byte, len(token.Data))
+	copy(token.Data, empty)
```
orenbm: what happens to `empty`? is this to clear the token from the memory or just from the `AccessToken` object? if it's the second one, why do we do this?

pkg/storage/access_token_file.go
```diff
@@ -1,25 +1,70 @@
...
+		err = os.MkdirAll(tokenDir, 755)
+		if err != nil {
+			// Do not specifying the directory in the error message from security reasons
+			return fmt.Errorf("error writing access token, reason: failed to create directory")
+		}
+	}
+
+	err = ioutil.WriteFile(token.TokenFilePath, token.Data, 0644)
+	if err != nil {
+		// Do not specifying the file path in the error message from security reasons
```
orenbm: do not specify

orenbm: for security reasons

pkg/storage/access_token_file.go
```diff
@@ -1,25 +1,70 @@
...
+
+	return nil
 }
 
 func (token *AccessTokenFile) Delete() (err error) {
-	return
-}
+	err = os.Remove(token.TokenFilePath)
+	if err != nil {
+		// Do not specifying the file path in the error message from security reasons
```
orenbm: Do not specify the file path in the error message for security reasons

cmd/sidecar/main.go
```diff
@@ -30,33 +30,34 @@ func main() {
 	tokenFilePath := flag.String("t", "/run/conjur/access-token",
 		"Path to Conjur access token")
 
-	configAuthn, err := authnConfig.NewFromEnv(clientCertPath, tokenFilePath)
+	// Create new Storage
+	configStorage, err := storageConfig.NewFromEnv(tokenFilePath)
```
orenbm: don't we want to read this as an env var instead? this can go inside `storage/config`

pkg/storage/access_token_file_test.go
```diff
@@ -0,0 +1,126 @@
...
+import (
+	stoargeConfig "github.com/cyberark/conjur-authn-k8s-client/pkg/storage/config"
+	. "github.com/smartystreets/goconvey/convey"
+	"os"
+	"reflect"
+	"testing"
+)
+
+func TestAccessTokenFile(t *testing.T) {
+	var config stoargeConfig.Config
```
orenbm: typo in `stoargeConfig`

pkg/storage/access_token_memory.go
nessiLahav: the first one - to clear the data from memory, do you think i should change here something ?

orenbm: how does it clear the data from memory? is `empty` handled by the garbage collector at the end of the method? that's the idea?

cmd/sidecar/main.go
```diff
@@ -30,33 +30,34 @@ func main() {
...
 	if err != nil {
 		errLogger.Printf(err.Error())
 		os.Exit(1)
 	}
 
-	// Create new Authenticator
-	authn, err := authenticator.New(*configAuthn)
+	storageHandler, err := storage.NewStorageHandler(*configStorage)
```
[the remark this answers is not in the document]
nessiLahav: I didn't want to change the secretConfig since i knew that you are working on it in your PR and maybe it should be merging with the storageConfig somehow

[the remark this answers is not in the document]
orenbm: cool so i'll take care of it

pkg/storage/access_token_memory.go
nessiLahav: empty - is an empty slice, and i copy this empty slice to Data

cmd/sidecar/main.go
```diff
@@ -125,6 +125,12 @@ func main() {
 				errLogger.Printf("Failure patching K8s Secrets: %s", err.Error())
 				return err
 			}
+
+			// Deleting Conjur access token
+			err = storageHandler.AccessToken.Delete()
```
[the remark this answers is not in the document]
nessiLahav: Thanks, i will add another delete action when existing the app

cmd/sidecar/main.go
```diff
@@ -142,6 +143,12 @@ func main() {
 
 	if err != nil {
 		errLogger.Printf("Backoff exhausted: %s", err.Error())
+		// Deleting the retrieved Conjur access token in case we got an error after retrieval.
+		// if the access token is already deleted the action should not fail
+		err = storageHandler.AccessToken.Delete()
+		if err != nil {
+			errLogger.Printf("￿￿Failed to delete access token: %s", err.Error())
```
sigalsax: I get weird characters here: "￿￿" what's that supposed to be?

cmd/sidecar/main.go
```diff
@@ -142,6 +143,12 @@ func main() {
 
 	if err != nil {
 		errLogger.Printf("Backoff exhausted: %s", err.Error())
+		// Deleting the retrieved Conjur access token in case we got an error after retrieval.
+		// if the access token is already deleted the action should not fail
+		err = storageHandler.AccessToken.Delete()
```
sigalsax: nice!

pkg/authenticator/authenticator.go
```diff
@@ -238,14 +240,7 @@ func (auth *Authenticator) ParseAuthenticationResponse(response []byte) error {
...
 
-	// InfoLogger.Printf("Writing token %v to shared volume ...", content)
-	// Create the directory if it doesn't exist
-	tokenDir := filepath.Dir(auth.Config.TokenFilePath)
-	if _, err := os.Stat(tokenDir); os.IsNotExist(err) {
-		os.MkdirAll(tokenDir, 755)
-	}
-
-	err = ioutil.WriteFile(auth.Config.TokenFilePath, content, 0644)
+	err = auth.AccessToken.Write(content)
```
sigalsax: why did we switch out `WriteFile` to just `Write`? How do we know the token file path here?

pkg/authenticator/config/config.go
```diff
@@ -77,6 +79,12 @@ func NewFromEnv(clientCertPath *string, tokenPath *string) (*Config, error) {
 		tokenRefreshTimeout = parsedTokenRefreshTimeout
 	}
 
+	clientCertPath := ClientCertPathDefault
+	// If CONJUR_CLIENT_CERT_PATH not configured take default value
+	if envVal := os.Getenv("CONJUR_CLIENT_CERT_PATH"); envVal != "" {
```
sigalsax: nice!

pkg/storage/access_token_file.go
```diff
@@ -1,25 +1,70 @@
...
+	tokenDir := filepath.Dir(token.TokenFilePath)
+	if _, err := os.Stat(tokenDir); os.IsNotExist(err) {
+		err = os.MkdirAll(tokenDir, 755)
+		if err != nil {
+			// Do not specify the directory in the error message for security reasons
+			return fmt.Errorf("error writing access token, reason: failed to create directory")
+		}
+	}
+
+	err = ioutil.WriteFile(token.TokenFilePath, token.Data, 0644)
```
sigalsax: why are we writing the access token twice? once in authn and once here in access_token_file?

pkg/authenticator/authenticator.go
sigalsax: why are we writing the access token twice? once here for authn and once for access token file?

pkg/storage/access_token_file_test.go
```diff
@@ -0,0 +1,149 @@
...
+	var tokenInFile, _ = NewAccessTokenFile(config)
+
+	Convey("Read", t, func() {
+		Convey("Returns true if Data output as expected", func() {
+			dataActual := []byte{'t', 'e', 's', 't'}
+			err := tokenInFile.Write(dataActual)
+			So(err, ShouldEqual, nil)
+
+			dataExpected, err := tokenInFile.Read()
+			eq := reflect.DeepEqual(dataActual, dataExpected)
```
sigalsax: when would dataActual and dataExpected ever be different? You just wrote dataActual and then immediately read it so why is this `So(eq, ShouldEqual, true)` a good indicator that our code works?

pkg/storage/access_token_file.go
```diff
@@ -1,25 +1,70 @@
...
+)
 
 type AccessTokenFile struct {
-	TokenFilePath	string
-	Data	[]byte
+	TokenFilePath string
+	Data          []byte
 }
 
 func NewAccessTokenFile(config config.Config) (token *AccessTokenFile, err error) {
```
sigalsax: We sure that this is a good name for this function? To me, `NewAccessTokenFile` represents a file and not the access token object that holds the data and the path. Can't we just leave it as NewAccessToken or do we have two AccessToken structs (1 for old flow with file path and 1 for new flow with token in memory?

pkg/storage/access_token_file_test.go
```diff
@@ -0,0 +1,149 @@
...
+
+		Convey("Returns error if Data is nil", func() {
+			tokenInFile.Data = nil
+			_, err := tokenInFile.Read()
+
+			So(err.Error(), ShouldEqual, "error reading access token, reason: data is empty")
+		})
+	})
+
+	Convey("Write", t, func() {
```
sigalsax: Don't we also want to check if the inverse - return an error if data input is null? I believe in code we raise an error

pkg/storage/access_token_file_test.go
```diff
@@ -0,0 +1,149 @@
...
+			err := tokenInFile.Write(dataActual)
+			So(err, ShouldEqual, nil)
+
+			// Check if file exits
+			_, err = os.Stat("/tmp/accessTokenFileWrite1")
+			So(err, ShouldEqual, nil)
+
+			// Second time
+			dataActual = []byte{'t', 'e', 's', 't', '2'}
+			err = tokenInFile.Write(dataActual)
```
sigalsax: Writing on same file will result in an override of the file contents or an appending ?

pkg/storage/access_token_file_test.go
```diff
@@ -0,0 +1,149 @@
...
+			err := tokenInFile.Write(dataActual)
+			So(err, ShouldEqual, nil)
+
+			_, err = tokenInFile.Read()
+			So(err, ShouldEqual, nil)
+
+			err = tokenInFile.Delete()
+			So(err, ShouldEqual, nil)
+
+			// Check that file is not exits
```
sigalsax: "check that file does not exist"

sigalsax: why wouldn't the file exist after delete func is run? Isn't delete just for deleting _content_ from access token?

PR #32, opened 2019-08-25
Update k8s secrets works e2e

orenbm:
Implemented the whole body of the design. Also merged the access-token refactor so the feature works now e2e.


cmd/sidecar/main.go
```diff
@@ -21,58 +19,41 @@ var infoLogger = log.InfoLogger
...
-	configStorage, err := storageConfig.NewFromEnv()
+	// Initialize configurations
+	authnConfig, err := authnConfigProvider.NewFromEnv()
 	if err != nil {
 		errLogger.Printf(err.Error())
 		os.Exit(1)
 	}
 
-	storageHandler, err := storage.NewStorageHandler(*configStorage)
+	secretsConfig, err := secretsConfigProvider.NewFromEnv()
```
nessiLahav: I think we should move this logic inside NewStorageHandler function

cmd/sidecar/main.go
```diff
@@ -145,9 +109,9 @@ func main() {
 		errLogger.Printf("Backoff exhausted: %s", err.Error())
 		// Deleting the retrieved Conjur access token in case we got an error after retrieval.
 		// if the access token is already deleted the action should not fail
-		err = storageHandler.AccessToken.Delete()
+		err = storageHandler.AccessTokenHandler.Delete()
 		if err != nil {
-			errLogger.Printf("￿￿Failed to delete access token: %s", err.Error())
+			errLogger.Printf("failed to delete access token: %s", err.Error())
```
nessiLahav: You changed in to lower case but above messages are in upper

pkg/secrets/handlers/secrets_handler_k8s.go
```diff
@@ -0,0 +1,130 @@
...
+	ConjurSecretsFetcher conjur.ConjurSecretsFetcherInterface
+	K8sSecretsHandler    k8s.K8sSecretsHandler
+}
+
+var errLogger = log.ErrorLogger
+
+func NewSecretHandlerK8sUseCase(secretsConfig secretsConfig.Config, AccessTokenHandler access_token.AccessTokenHandler) (SecretsHandler *SecretsHandlerK8sUseCase, err error) {
+	k8sSecretsHandler, err := k8s.New(secretsConfig)
+	if err != nil {
+		errLogger.Printf(err.Error())
```
nessiLahav: Add error message and change `os.exit `to `return err` for to main to decide if to exit or not

pkg/secrets/handlers/secrets_handler_k8s.go
```diff
@@ -0,0 +1,130 @@
...
+	var conjurSecretsFetcher conjur.ConjurSecretsFetcher
+
+	return &SecretsHandlerK8sUseCase{
+		AccessTokenHandler:   AccessTokenHandler,
+		ConjurSecretsFetcher: conjurSecretsFetcher,
+		K8sSecretsHandler:    *k8sSecretsHandler,
+	}, nil
+}
+
+func (secretsHandlerK8sUseCase *SecretsHandlerK8sUseCase) HandleSecrets() error {
```
nessiLahav: change SecretsHandlerK8sUseCase(without *) it to copy by value since we are not changing anything in the struct

cmd/sidecar/main.go
orenbm: done. it may change in the future to one config and then we'll get it from main anyway.

pkg/secrets/handlers/secrets_handler_k8s.go
```diff
@@ -0,0 +1,130 @@
...
+		errLogger.Printf("Failure retrieving access token: %s", err.Error())
+		return err
+	}
+
+	variableIDs, err := getVariableIDsToRetrieve(k8sSecretsMap.PathMap)
+	if err != nil {
+		return fmt.Errorf("Error parsing Conjur variable ids: %s", err)
+	}
+
+	retrievedConjurSecrets, err := retrieveRequiredConjurSecrets(
```
nessiLahav: 1. Why this function is not a part of the ConjurSecretsFetcher, i expected to syntax like: `retrievedConjurSecrets, err := secretsHandlerK8sUseCase.ConjurSecretsFetcher.retrieveRequiredConjurSecrets(access_token, variableIDs)`
2. Do you clear access_token & ConjurSecretsFetcher after the use?

pkg/secrets/handlers/secrets_handler_none.go
```diff
@@ -0,0 +1,7 @@
+package handlers
+
+type SecretHandlerNoneUseCase struct{}
+
+func (secretHandlerNone *SecretHandlerNoneUseCase) HandleSecrets() error {
```
nessiLahav: delete *

pkg/secrets/handlers/secrets_handler_k8s.go
orenbm: 1. now that i look at the implementation of the method it's quite redundant. probably had more meat in it and got thinner. I'll remove it.
2. should i clear `access_token`? it is cleared in `main`. what do you mean by clearing `ConjurSecretsFetcher`?

pkg/secrets/k8s/k8s_secrets.go
```diff
@@ -0,0 +1,218 @@
...
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+	"strings"
+)
+
+// K8sSecretsHandler contains the configuration and client
+// for the authentication connection to Conjur
+type K8sSecretsHandler struct {
+	Config      secretsConfig.Config
+	AccessToken []byte
```
nessiLahav: delete AccessToken

pkg/secrets/k8s/k8s_secrets.go
```diff
@@ -0,0 +1,218 @@
...
+
+	kubeClient, err := kubernetes.NewForConfig(kubeConfig)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create Kubernetes client: %s", err)
+	}
+	// return a K8s client
+	return kubeClient, err
+}
+
+func (secrets *K8sSecretsHandler) RetrieveK8sSecrets() (*K8sSecretsMap, error) {
```
nessiLahav: delete *

pkg/secrets/handlers/secrets_handler_k8s.go
nessiLahav: 2. edited the comment, about access_token i think when we called read() we returned a copy of the data, if this is the case we should decide if to change the interface to return an pointer to data, or it will be the responsibility of the called to clear the data

nessiLahav: ignore 2 it got clean i will add UT that test it, just test the same issue regarding retrievedConjurSecrets

pkg/access_token/access_token_memory_test.go
```diff
@@ -14,7 +14,7 @@ type ProxyHandlerTokenMemory struct {
 func TestAccessTokenMemory(t *testing.T) {
 	var config stoargeConfig.Config
 	config.StoreType = stoargeConfig.None
```
sigalsax: spelling mistake :l "stoargeConfig"

PR #34, opened 2019-08-26
fix unstable ut

nessiLahav:
fix unstable ut
- sort maps and slices before comparison

### DIFF
--- a/pkg/secrets/handlers/secrets_handler_k8s_test.go
+++ b/pkg/secrets/handlers/secrets_handler_k8s_test.go
@@ -17,7 +17,7 @@ func TestSecretsHandlerK8sUseCase(t *testing.T) {
 			variableIDsExpected := []string{"account/var_path1", "account/var_path2"}
 			variableIDsActual, err := getVariableIDsToRetrieve(m)
 
-			// Sort actual and expected, because output order can be change
+			// Sort actual and expected, because output order can change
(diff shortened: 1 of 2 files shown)


pkg/secrets/handlers/secrets_handler_k8s_test.go
```diff
@@ -16,6 +17,9 @@ func TestSecretsHandlerK8sUseCase(t *testing.T) {
 			variableIDsExpected := []string{"account/var_path1", "account/var_path2"}
 			variableIDsActual, err := getVariableIDsToRetrieve(m)
 
+			// Sort actual and expected, because output order can be change
```
orenbm: can change

pkg/secrets/k8s/k8s_secrets_test.go
```diff
@@ -12,11 +15,22 @@ func TestKubernetesSecrets(t *testing.T) {
 			m := make(map[string][]byte)
 			m["user"] = []byte("dummy_user")
 			m["password"] = []byte("dummy_password")
-			stringDataEntryExpected := `{"stringData":{"user":"dummy_user","password":"dummy_password"}}`
+			stringDataEntryExpected := `{"stringData":{"user":"dummy_user","password":"dummy_password","address":"dummy_address"}}`
```
orenbm: i don't think we need address

PR #36, opened 2019-08-26
Add UTs to secrets handler

sigalsax:
- [ ] adds UTs for secret handler
 - [ ] refactors existing go tests to be more descriptive of what they are testing

Connected with #38

### DIFF
--- a/pkg/access_token/access_token_file_test.go
+++ b/pkg/access_token/access_token_file_test.go
@@ -19,138 +19,200 @@ func TestAccessTokenFile(t *testing.T) {
 	var tokenInFile, _ = NewAccessTokenFile(config)
 
 	Convey("Read", t, func() {
-		Convey("Returns true if Data output as expected", func() {
+
+		Convey("Given an access token with data", func() {
 			dataActual := []byte{'t', 'e', 's', 't'}
 			err := tokenInFile.Write(dataActual)
-			So(err, ShouldEqual, nil)
 
-			dataExpected, err := tokenInFile.Read()
-			eq := reflect.DeepEqual(dataActual, dataExpected)
+			Convey("Finishes without raising an error", func() {
+				So(err, ShouldEqual, nil)
+			})
 
-			So(err, ShouldEqual, nil)
-			So(eq, ShouldEqual, true)
+			Convey("When running Read method", func() {
+				dataExpected, err := tokenInFile.Read()
+
+				Convey("Finishes without raising an error", func() {
+					So(err, ShouldEqual, nil)
+				})
+
+				Convey("Returns the data that was written", func() {
+					eq := reflect.DeepEqual(dataActual, dataExpected)
+					So(eq, ShouldEqual, true)
+				})
+			})
 		})
 
-		Convey("Returns error if Data is nil", func() {
+		Convey("Given an access token's data is empty", func () {
 			tokenInFile.Data = nil
-			_, err := tokenInFile.Read()
 
-			So(err.Error(), ShouldEqual, "error reading access token, reason: data is empty")
+			Convey("When the token data is empty and running the Read method", func() {
+				_, err := tokenInFile.Read()
+
+				Convey("Raises an error that the data is empty", func() {
+					So(err.Error(), ShouldEqual, "error reading access token, reason: data is empty")
+				})
+			})
 		})
 	})
 
 	Convey("Write", t, func() {
-		Convey("Returns no error if Data input is not nil", func() {
+
+		Convey("Given an access token with data", func() {
 			dataActual := []byte{'t', 'e', 's', 't'}
-			tokenInFile.TokenFilePath = "/tmp/accessTokenFileWrite1"
-			err := tokenInFile.Write(dataActual)
 
-			So(err, ShouldEqual, nil)
-			// Check if file exits
-			_, err = os.Stat("/tmp/accessTokenFileWrite1")
-			So(err, ShouldEqual, nil)
+			Convey("When running the Write method, writes the access token to a file in the path defined", func() {
+				tokenInFile.TokenFilePath = "/tmp/accessTokenFileWrite1"
+				err := tokenInFile.Write(dataActual)
+
+				Convey("Finishes without raising an error", func() {
+					So(err, ShouldEqual, nil)
+					// Check if file exists
+					_, err = os.Stat("/tmp/accessTokenFileWrite1")
+					So(err, ShouldEqual, nil)
+				})
+			})
 		})
 
-		Convey("Returns error if Data input is nil", func() {
-			err := tokenInFile.Write(nil)
+		Convey("Given an access token without data", func() {
 
-			So(err.Error(), ShouldEqual, "error writing access token, reason: data is empty")
+			Convey("When running the Write method and attempts to write the access token", func() {
+				err := tokenInFile.Write(nil)
+
+				Convey("Raises an error that the access token data is empty", func() {
+					So(err.Error(), ShouldEqual, "error writing access token, reason: data is empty")
+				})
+			})
 		})
 
-		Convey("Returns no error if file already exists", func() {
+		Convey("Given access token with data and a file location defined", func() {
 			dataActual := []byte{'t', 'e', 's', 't'}
 			tokenInFile.TokenFilePath = "/tmp/accessTokenFileWrite2"
-			err := tokenInFile.Write(dataActual)
-			So(err, ShouldEqual, nil)
 
-			// Check if file exits
-			_, err = os.Stat("/tmp/accessTokenFileWrite1")
-			So(err, ShouldEqual, nil)
+			Convey("When running the Write method", func() {
+				err := tokenInFile.Write(dataActual)
 
-			// Second time
-			dataActual = []byte{'t', 'e', 's', 't', '2'}
-			err = tokenInFile.Write(dataActual)
-			So(err, ShouldEqual, nil)
+				Convey("Finishes without raising an error", func() {
+					So(err, ShouldEqual, nil)
 
-			// Check if file exits
-			_, err = os.Stat("/tmp/accessTokenFileWrite1")
-			So(err, ShouldEqual, nil)
+					// Check if file exists
+					_, err = os.Stat("/tmp/accessTokenFileWrite2")
+					So(err, ShouldEqual, nil)
+				})
+			})
 
-			// Test we are reading the new Data
-			dataExpected, err := tokenInFile.Read()
-			eq := reflect.DeepEqual(dataActual, dataExpected)
+			Convey("When running the Write method a second time", func() {
+				dataActual = []byte{'t', 'e', 's', 't', '2'}
+				err := tokenInFile.Write(dataActual)
+				So(err, ShouldEqual, nil)
 
-			So(err, ShouldEqual, nil)
-			So(eq, ShouldEqual, true)
+				// Check if file exits
+				_, err = os.Stat("/tmp/accessTokenFileWrite2")
+				So(err, ShouldEqual, nil)
+
+				Convey("Raises no error and the data should be written to the file as expected", func() {
+					// Test we are reading the new Data
+					dataExpected, err := tokenInFile.Read()
+					eq := reflect.DeepEqual(dataActual, dataExpected)
+
+					So(err, ShouldEqual, nil)
+					So(eq, ShouldEqual, true)
+				})
+
+			})
 		})
 	})
 
 	Convey("Delete", t, func() {
-		Convey("Returns no error after read write and delete", func() {
+
+		Convey("Given access token with data", func() {
 			dataActual := []byte{'t', 'e', 's', 't'}
-			tokenInFile.TokenFilePath = "/tmp/accessTokenFileDel1"
-			err := tokenInFile.Write(dataActual)
-			So(err, ShouldEqual, nil)
 
-			dataFromRead, err := tokenInFile.Read()
-			So(err, ShouldEqual, nil)
+			Convey("When running the Write method", func() {
+				tokenInFile.TokenFilePath = "/tmp/accessTokenFileDel1"
+				err := tokenInFile.Write(dataActual)
 
-			err = tokenInFile.Delete()
-			So(err, ShouldEqual, nil)
+				Convey("Finishes without raising an error", func() {
+					So(err, ShouldEqual, nil)
+				})
 
-			// Check that file does not exits
-			_, err = os.Stat("/tmp/accessTokenFileDel1")
-			So(err, ShouldNotEqual, nil)
+				Convey("When running the Read method", func() {
+					dataFromRead, err := tokenInFile.Read()
 
-			// Both input & output arrays should be cleared from memory
-			empty := make([]byte, len(dataActual))
-			eq := reflect.DeepEqual(dataActual, empty)
-			So(eq, ShouldEqual, true)
-			eq = reflect.DeepEqual(dataFromRead, empty)
-			So(eq, ShouldEqual, true)
+					Convey("Finishes without raising an error", func() {
+						So(err, ShouldEqual, nil)
+					})
+
+					Convey("When running the Delete method", func() {
+						err = tokenInFile.Delete()
+
+						Convey("Finishes without raising an error", func() {
+							So(err, ShouldEqual, nil)
+						})
+
+						Convey("Properly deletes the file", func() {
+							_, err = os.Stat("/tmp/accessTokenFileDel1")
+							So(err, ShouldNotEqual, nil)
+						})
+
+						Convey("Properly clears all data from memory", func() {
+							// Both input & output arrays should be cleared from memory
+							empty := make([]byte, len(dataActual))
+							eq := reflect.DeepEqual(dataActual, empty)
+							So(eq, ShouldEqual, true)
+							eq = reflect.DeepEqual(dataFromRead, empty)
+							So(eq, ShouldEqual, true)
+						})
+					})
+				})
+			})
 		})
 
-		Convey("Returns no error if Data input is nil", func() {
+		Convey("Given an access token with no data saved to a file", func() {
 			tokenInFile.Data = nil
 			os.Create("/tmp/accessTokenFileDel2")
 			tokenInFile.TokenFilePath = "/tmp/accessTokenFileDel2"
-			err := tokenInFile.Delete()
-			So(err, ShouldEqual, nil)
 
-			// Check that file is not exits
-			_, err = os.Stat("/tmp/accessTokenFileDel2")
-			So(err, ShouldNotEqual, nil)
+			Convey("When running the Delete method", func() {
+				err := tokenInFile.Delete()
+
+				Convey("Deletes the file and no errors are raised", func() {
+					So(err, ShouldEqual, nil)
+
+					// Check that file does not exist
+					_, err = os.Stat("/tmp/accessTokenFileDel2")
+					So(err, ShouldNotEqual, nil)
+				})
+
+				Convey("When running the Delete method again on the same file", func() {
+					err = tokenInFile.Delete()
+
+					Convey("Finishes with an error deleting the access token", func() {
+						So(err.Error(), ShouldEqual, "error deleting access token")
+					})
+				})
+			})
 		})
 
-		Convey("Returns error if file already deleted", func() {
-			tokenInFile.Data = nil
-			os.Create("/tmp/accessTokenFileDel3")
-			tokenInFile.TokenFilePath = "/tmp/accessTokenFileDel3"
-			err := tokenInFile.Delete()
-			So(err, ShouldEqual, nil)
+		Convey("Given two instances of the accessTokenHandler interface", func() {
 
-			// Check that file is not exits
-			err = tokenInFile.Delete()
-			So(err.Error(), ShouldEqual, "error deleting access token")
-		})
+			Convey("Finishes without raising errors and data has been deleted", func() {
+				// Write Data to source interface
+				dataActual := []byte{'t', 'e', 's', 't'}
+				tokenInFile.Write(dataActual)
 
-		Convey("Returns no error if delete data from proxy struct is as expected", func() {
-			// Write Data to source interface
-			dataActual := []byte{'t', 'e', 's', 't'}
-			tokenInFile.Write(dataActual)
+				// Set proxy struct with source interface
+				var proxyStruct ProxyHandlerTokenFile
+				proxyStruct.AccessToken = tokenInFile
 
-			// Set proxy struct with source interface
-			var proxyStruct ProxyHandlerTokenFile
-			proxyStruct.AccessToken = tokenInFile
+				// Delete access token from proxy
+				err := proxyStruct.AccessToken.Delete()
+				So(err, ShouldEqual, nil)
 
-			// Delete access token from proxy
-			err := proxyStruct.AccessToken.Delete()
-			So(err, ShouldEqual, nil)
-
-			// Data in source interface should be deleted
-			dataExpected, err := tokenInFile.Read()
-			So(err.Error(), ShouldEqual, "error reading access token, reason: data is empty")
-			So(dataExpected, ShouldEqual, nil)
+				// Data in source interface should be deleted
+				dataExpected, err := tokenInFile.Read()
+				So(err.Error(), ShouldEqual, "error reading access token, reason: data is empty")
+				So(dataExpected, ShouldEqual, nil)
+			})
 		})
 	})
 }
--- a/pkg/access_token/access_token_file_test.go
+++ b/pkg/access_token/access_token_file_test.go
@@ -42,7 +42,7 @@ func TestAccessTokenFile(t *testing.T) {
 			})
 		})
 
-		Convey("Given an access token's data is empty", func () {
+		Convey("Given an access token's data is empty", func() {
 			tokenInFile.Data = nil
 
 			Convey("When running the Read method", func() {
@@ -57,28 +57,45 @@ func TestAccessTokenFile(t *testing.T) {
 
 	Convey("Write", t, func() {
 
-		Convey("Given an access token with data", func() {
+		Convey("Given an access token with data and a defined file location", func() {
 			dataActual := []byte{'t', 'e', 's', 't'}
+			tokenInFile.TokenFilePath = "/tmp/accessTokenFileWrite1"
 
 			Convey("When running the Write method", func() {
-				tokenInFile.TokenFilePath = "/tmp/accessTokenFileWrite1"
 				err := tokenInFile.Write(dataActual)
 
 				Convey("Finishes without raising an error", func() {
 					So(err, ShouldEqual, nil)
 				})
 
-				Convey("Checks that the file exists in the path defined", func (){
-					// Check if file exists
+				Convey("Checks that the file exists in the path defined", func() {
 					_, err = os.Stat("/tmp/accessTokenFileWrite1")
 					So(err, ShouldEqual, nil)
 				})
 
-				Convey("When running Read method", func() {
+				Convey("And the data was read", func() {
 					dataExpected, _ := tokenInFile.Read()
 
 					// Confirm data was written
 					Convey("Returns the data the was written to the file", func() {
+						eq := reflect.DeepEqual(dataActual, dataExpected)
+						So(eq, ShouldEqual, true)
+					})
+				})
+
+				Convey("When running the Write method a second time", func() {
+					dataActual = []byte{'t', 'e', 's', 't', '2'}
+					err := tokenInFile.Write(dataActual)
+
+					Convey("Creates the file without raising an error", func() {
+						// Check if file exists
+						_, err = os.Stat("/tmp/accessTokenFileWrite1")
+						So(err, ShouldEqual, nil)
+					})
+
+					Convey("Writes the data to the file", func() {
+						// TODO: read the content with `os` methods (not with `tokenInFile`)
+						dataExpected, _ := tokenInFile.Read()
 						eq := reflect.DeepEqual(dataActual, dataExpected)
 						So(eq, ShouldEqual, true)
 					})
@@ -96,42 +113,6 @@ func TestAccessTokenFile(t *testing.T) {
 				})
 			})
 		})
-
-		Convey("Given access token with data and a file location defined", func() {
-			dataActual := []byte{'t', 'e', 's', 't'}
-			tokenInFile.TokenFilePath = "/tmp/accessTokenFileWrite2"
-
-			Convey("When running the Write method", func() {
-				err := tokenInFile.Write(dataActual)
-
-				Convey("Writes the access token without raising an error", func() {
-					So(err, ShouldEqual, nil)
-				})
-
-				Convey("Creates with file without raising an error", func() {
-					// Check if file exists
-					_, err = os.Stat("/tmp/accessTokenFileWrite2")
-					So(err, ShouldEqual, nil)
-				})
-			})
-
-			Convey("When running the Write method a second time", func() {
-				dataActual = []byte{'t', 'e', 's', 't', '2'}
-				err := tokenInFile.Write(dataActual)
-
-				Convey("Creates with file without raising an error", func() {
-					// Check if file exists
-					_, err = os.Stat("/tmp/accessTokenFileWrite2")
-					So(err, ShouldEqual, nil)
-				})
-
-				Convey("Writes the data the file as expected", func() {
-					dataExpected, _ := tokenInFile.Read()
-					eq := reflect.DeepEqual(dataActual, dataExpected)
-					So(eq, ShouldEqual, true)
-				})
-			})
-		})
 	})
 
 	Convey("Delete", t, func() {
@@ -139,15 +120,13 @@ func TestAccessTokenFile(t *testing.T) {
 		Convey("Given access token with data", func() {
 			dataActual := []byte{'t', 'e', 's', 't'}
 
-			Convey("When running the Write method", func() {
+			Convey("And the data was written successfully", func() {
 				tokenInFile.TokenFilePath = "/tmp/accessTokenFileDel1"
 				err := tokenInFile.Write(dataActual)
+				So(err, ShouldEqual, nil)
 
-				Convey("Finishes without raising an error", func() {
-					So(err, ShouldEqual, nil)
-				})
-
-				Convey("When running the Read method", func() {
+				// Read is added here because we want to check later that the contents were deleted from memory successfully
+				Convey("and the data was read successfully", func() {
 					dataFromRead, err := tokenInFile.Read()
 
 					Convey("Finishes without raising an error", func() {
@@ -198,7 +177,7 @@ func TestAccessTokenFile(t *testing.T) {
 				Convey("When running the Delete method again on the same file", func() {
 					err = tokenInFile.Delete()
 
-					Convey("Finishes with an error deleting the access token", func() {
+					Convey("Finishes with proper error", func() {
 						So(err.Error(), ShouldEqual, "error deleting access token")
 					})
 				})
@@ -210,7 +189,7 @@ func TestAccessTokenFile(t *testing.T) {
 			dataActual := []byte{'t', 'e', 's', 't'}
 			tokenInFile.Write(dataActual)
 
-			Convey ("When setting token file location in proxy struct", func(){
+			Convey("When setting token file location in proxy struct", func() {
 				// Set proxy struct with source interface
 				var proxyStruct ProxyHandlerTokenFile
 				proxyStruct.AccessToken = tokenInFile
--- a/pkg/access_token/access_token_file_test.go
+++ b/pkg/access_token/access_token_file_test.go
@@ -68,14 +68,23 @@ func TestAccessTokenFile(t *testing.T) {
 					So(err, ShouldEqual, nil)
 				})
 
-				Convey("Writes the access token to a file in the path defined", func (){
+				Convey("Checks that the file exists in the path defined", func (){
 					// Check if file exists
 					_, err = os.Stat("/tmp/accessTokenFileWrite1")
 					So(err, ShouldEqual, nil)
 				})
+
+				Convey("When running Read method", func() {
+					dataExpected, _ := tokenInFile.Read()
+
+					// Confirm data was written
+					Convey("Returns the data the was written to the file", func() {
+						eq := reflect.DeepEqual(dataActual, dataExpected)
+						So(eq, ShouldEqual, true)
+					})
+				})
 			})
 		})
-
 
 		Convey("Given an access token without data", func() {
 
--- a/pkg/access_token/access_token_file_test.go
+++ b/pkg/access_token/access_token_file_test.go
@@ -45,7 +45,7 @@ func TestAccessTokenFile(t *testing.T) {
 		Convey("Given an access token's data is empty", func () {
 			tokenInFile.Data = nil
 
-			Convey("When the token data is empty and running the Read method", func() {
+			Convey("When running the Read method", func() {
 				_, err := tokenInFile.Read()
 
 				Convey("Raises an error that the data is empty", func() {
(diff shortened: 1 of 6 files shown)


pkg/secrets/handlers/secrets_handler_k8s_test.go
```diff
@@ -29,4 +30,35 @@ func TestSecretsHandlerK8sUseCase(t *testing.T) {
 			So(err.Error(), ShouldEqual, "error map should not be empty")
 		})
 	})
+
+	Convey("updateK8sSecretsMapWithConjurSecrets", t, func () {
+		secret := []byte{'s', 'u', 'p', 'e', 'r'}
+		conjurSecrets := make(map[string][]byte)
+		conjurSecrets["account:variable:allowed/username"] = secret
+
+		newDataEntriesMap := make(map[string][]byte)
```
orenbm: please change this to `dataEntriesMap`

pkg/secrets/handlers/secrets_handler_k8s_test.go
```diff
@@ -29,4 +30,35 @@ func TestSecretsHandlerK8sUseCase(t *testing.T) {
...
+		Convey("Returns true if K8sSecretsMap's variable ID is replaced with corresponding secret value", func() {
+			So(err, ShouldEqual, nil)
+			So(eq, ShouldEqual, true)
+		})
+
+		Convey("Returns true if Conjur secret byte array is cleared from memory", func() {
+			// secret value should be cleared from memory after use
+			empty := make([]byte, len(secret))
+			eq = reflect.DeepEqual(conjurSecrets["account:variable:allowed/username"], empty)
+			So(eq, ShouldEqual, true)
```
orenbm: does this actually work? did you change the code where we clear the secret from memory?

pkg/secrets/handlers/secrets_handler_k8s_test.go
```diff
@@ -29,4 +30,35 @@ func TestSecretsHandlerK8sUseCase(t *testing.T) {
...
+		k8sSecretsMap["mysecret"] = newDataEntriesMap
+
+		pathMap := make(map[string]string)
+		pathMap["allowed/username"] = "mysecret:username"
+
+		k8sSecretsStruct := k8s.K8sSecretsMap{k8sSecretsMap, pathMap}
+		err := updateK8sSecretsMapWithConjurSecrets(&k8sSecretsStruct, conjurSecrets)
+
+		eq := reflect.DeepEqual(k8sSecretsStruct.K8sSecrets["mysecret"]["username"], secret)
+		Convey("Returns true if K8sSecretsMap's variable ID is replaced with corresponding secret value", func() {
```
orenbm: the message here is not what we're looking for. We need to write here the expected behaviour. For example:
`Convey("Replaces the Variable IDs in k8sSecretsMap to their corresponding secret values without errors"...`

This is actually a mistake that nessi did in his UTs (you can see how he got the idea from `authenticator_test.go`) so please go over all the `Convey`s in our code and change the text to be self-explanatory. This will also help you understand what we test in each UT.

pkg/secrets/handlers/secrets_handler_k8s_test.go
```diff
@@ -29,4 +30,35 @@ func TestSecretsHandlerK8sUseCase(t *testing.T) {
...
+			So(err, ShouldEqual, nil)
+			So(eq, ShouldEqual, true)
+		})
+
+		Convey("Returns true if Conjur secret byte array is cleared from memory", func() {
```
orenbm: Regarding the 2 tests separation - i think this is quite good but I would add another Convey that wraps them. For example `Convey("One K8s secret with one Conjur secret"...` or something in that manner. It will be easier to add more scenarios to this test. please see the example here: http://goconvey.co/ (under Expressive DSL)

pkg/secrets/handlers/secrets_handler_k8s_test.go
```diff
@@ -29,4 +30,35 @@ func TestSecretsHandlerK8sUseCase(t *testing.T) {
...
+		k8sSecretsMap := make(map[string]map[string][]byte)
+		k8sSecretsMap["mysecret"] = newDataEntriesMap
+
+		pathMap := make(map[string]string)
+		pathMap["allowed/username"] = "mysecret:username"
+
+		k8sSecretsStruct := k8s.K8sSecretsMap{k8sSecretsMap, pathMap}
+		err := updateK8sSecretsMapWithConjurSecrets(&k8sSecretsStruct, conjurSecrets)
+
+		eq := reflect.DeepEqual(k8sSecretsStruct.K8sSecrets["mysecret"]["username"], secret)
```
orenbm: I would move this inside the convey

pkg/secrets/handlers/secrets_handler_k8s_test.go
sigalsax: Given When Then seems to be the structure they use. I could say:
`Given one K8s secret with one Conjur secret then the Variable ID for the Conjur secret is replaced in k8sSecretsMap with their corresponding secret value`

orenbm: I think we should go with `Given` -> `it` like we have in ruby UTs.

take this for an example:
```
Convey("updateK8sSecretsMapWithConjurSecrets", t, func () {
	Convey("Given one K8s secret with one Conjur secret", func() {
		// define all mocks

		err := updateK8sSecretsMapWithConjurSecrets(&k8sSecretsStruct, conjurSecrets)

		Convey("Finishes without raising an error", func() {
			So(err, ShouldEqual, nil)
		})

		Convey("Replaces the Variable ID with their corresponding secret value in k8sSecretsMap", func() {
			eq := reflect.DeepEqual(k8sSecretsStruct.K8sSecrets["mysecret"]["username"], secret)
			So(eq, ShouldEqual, true)
		})

		Convey("Clears the Conjur secret byte array from memory", func() {
			// secret value should be cleared from memory after use
			empty := make([]byte, len(secret))
			eq = reflect.DeepEqual(conjurSecrets["account:variable:allowed/username"], empty)
			So(eq, ShouldEqual, true)
		})
	})
})
```

orenbm: each thing we test has its own `Convey` section so we pin point test failures. this way once a test is failing we don't need to further investigate *what* failed in it. when you write the "it" convey imagine the sentence starts with "it"

sigalsax: Is `Given` what is passed as a parm in the function? or just set the context?

orenbm: in general `given` is everything that happens before what you are actually testing. it is things you assume are working and you're not testing them now. they are given to you.

pkg/access_token/access_token_file_test.go
```diff
@@ -19,138 +19,200 @@ func TestAccessTokenFile(t *testing.T) {
...
+			})
 		})
 
-		Convey("Returns error if Data is nil", func() {
+		Convey("Given an access token's data is empty", func () {
 			tokenInFile.Data = nil
-			_, err := tokenInFile.Read()
 
-			So(err.Error(), ShouldEqual, "error reading access token, reason: data is empty")
+			Convey("When the token data is empty and running the Read method", func() {
```
orenbm: This can change to `When running Read method`. the prefix is already in the `given` clause

pkg/access_token/access_token_file_test.go
```diff
@@ -19,138 +19,200 @@ func TestAccessTokenFile(t *testing.T) {
...
+		Convey("Given an access token with data", func() {
 			dataActual := []byte{'t', 'e', 's', 't'}
-			tokenInFile.TokenFilePath = "/tmp/accessTokenFileWrite1"
-			err := tokenInFile.Write(dataActual)
 
-			So(err, ShouldEqual, nil)
-			// Check if file exits
-			_, err = os.Stat("/tmp/accessTokenFileWrite1")
-			So(err, ShouldEqual, nil)
+			Convey("When running the Write method, writes the access token to a file in the path defined", func() {
```
orenbm: i think this should be:
```
Convey("When running the Write method, ", func() {
	tokenInFile.TokenFilePath = "/tmp/accessTokenFileWrite1"
	err := tokenInFile.Write(dataActual)

	Convey("Finishes without raising an error", func() {
		So(err, ShouldEqual, nil)
	})

	Convey("Writes the access token to a file in the path defined", func() {
		_, err = os.Stat("/tmp/accessTokenFileWrite1")
		So(err, ShouldEqual, nil)
	})
})
```

pkg/access_token/access_token_file_test.go
```diff
@@ -19,138 +19,200 @@ func TestAccessTokenFile(t *testing.T) {
...
+				})
+			})
 		})
 
-		Convey("Returns error if Data input is nil", func() {
-			err := tokenInFile.Write(nil)
+		Convey("Given an access token without data", func() {
 
-			So(err.Error(), ShouldEqual, "error writing access token, reason: data is empty")
+			Convey("When running the Write method and attempts to write the access token", func() {
```
orenbm: "and attempts to write the access token" is redundant

pkg/access_token/access_token_file_test.go
```diff
@@ -19,138 +19,200 @@ func TestAccessTokenFile(t *testing.T) {
...
-			err := tokenInFile.Write(dataActual)
-			So(err, ShouldEqual, nil)
 
-			// Check if file exits
-			_, err = os.Stat("/tmp/accessTokenFileWrite1")
-			So(err, ShouldEqual, nil)
+			Convey("When running the Write method", func() {
+				err := tokenInFile.Write(dataActual)
+
+				Convey("Finishes without raising an error", func() {
```
orenbm: please split to 2 conveys

pkg/access_token/access_token_file_test.go
```diff
@@ -19,138 +19,200 @@ func TestAccessTokenFile(t *testing.T) {
...
+				})
+			})
 
-			// Check if file exits
-			_, err = os.Stat("/tmp/accessTokenFileWrite1")
-			So(err, ShouldEqual, nil)
+			Convey("When running the Write method a second time", func() {
+				dataActual = []byte{'t', 'e', 's', 't', '2'}
+				err := tokenInFile.Write(dataActual)
+				So(err, ShouldEqual, nil)
```
orenbm: this line is duplicated and can be removed. you're running it inside the next convey

pkg/access_token/access_token_file_test.go
```diff
@@ -19,138 +19,200 @@ func TestAccessTokenFile(t *testing.T) {
...
-			// Test we are reading the new Data
-			dataExpected, err := tokenInFile.Read()
-			eq := reflect.DeepEqual(dataActual, dataExpected)
+				// Check if file exits
+				_, err = os.Stat("/tmp/accessTokenFileWrite2")
+				So(err, ShouldEqual, nil)
 
-			So(err, ShouldEqual, nil)
-			So(eq, ShouldEqual, true)
+				Convey("Raises no error and the data should be written to the file as expected", func() {
```
orenbm: please split to 2 conveys

pkg/access_token/access_token_file_test.go
```diff
@@ -19,138 +19,200 @@ func TestAccessTokenFile(t *testing.T) {
...
+			Convey("When running the Write method a second time", func() {
+				dataActual = []byte{'t', 'e', 's', 't', '2'}
+				err := tokenInFile.Write(dataActual)
+				So(err, ShouldEqual, nil)
 
-			// Test we are reading the new Data
-			dataExpected, err := tokenInFile.Read()
-			eq := reflect.DeepEqual(dataActual, dataExpected)
+				// Check if file exits
+				_, err = os.Stat("/tmp/accessTokenFileWrite2")
```
orenbm: should this go inside the convey?

pkg/access_token/access_token_file_test.go
```diff
@@ -19,138 +19,200 @@ func TestAccessTokenFile(t *testing.T) {
...
+
+					So(err, ShouldEqual, nil)
+					So(eq, ShouldEqual, true)
+				})
+
+			})
 		})
 	})
 
 	Convey("Delete", t, func() {
```
orenbm: this test is great!

orenbm: beautifully written and easy to read

pkg/access_token/access_token_file_test.go
```diff
@@ -19,138 +19,200 @@ func TestAccessTokenFile(t *testing.T) {
...
+
+			Convey("Finishes without raising errors and data has been deleted", func() {
+				// Write Data to source interface
+				dataActual := []byte{'t', 'e', 's', 't'}
+				tokenInFile.Write(dataActual)
 
-			// Set proxy struct with source interface
-			var proxyStruct ProxyHandlerTokenFile
-			proxyStruct.AccessToken = tokenInFile
+				// Set proxy struct with source interface
```
orenbm: can this be in a `when` convey?

pkg/access_token/access_token_file_test.go
```diff
@@ -19,138 +19,200 @@ func TestAccessTokenFile(t *testing.T) {
...
+				var proxyStruct ProxyHandlerTokenFile
+				proxyStruct.AccessToken = tokenInFile
 
-			// Delete access token from proxy
-			err := proxyStruct.AccessToken.Delete()
-			So(err, ShouldEqual, nil)
+				// Delete access token from proxy
```
orenbm: same here. anywhere you have a comment explaining the step is a good point to add a convey

pkg/access_token/access_token_file_test.go
```diff
@@ -19,138 +19,200 @@ func TestAccessTokenFile(t *testing.T) {
...
+				err := proxyStruct.AccessToken.Delete()
+				So(err, ShouldEqual, nil)
 
-			// Data in source interface should be deleted
-			dataExpected, err := tokenInFile.Read()
-			So(err.Error(), ShouldEqual, "error reading access token, reason: data is empty")
-			So(dataExpected, ShouldEqual, nil)
+				// Data in source interface should be deleted
```
orenbm: then

pkg/access_token/access_token_memory_test.go
```diff
@@ -16,84 +16,124 @@ func TestAccessTokenMemory(t *testing.T) {
...
 
 	Convey("Write", t, func() {
-		Convey("Returns no error if Data input is not nil", func() {
+
+		Convey("Given an access token with data", func() {
 			dataActual := []byte{'t', 'e', 's', 't'}
-			err := tokenInMemory.Write(dataActual)
 
-			So(err, ShouldEqual, nil)
+			Convey("Writes the access token to memory without raising an error", func() {
```
orenbm: can we verify it is actually written to memory? here you just verify there is no error

pkg/access_token/access_token_memory_test.go
```diff
@@ -16,84 +16,124 @@ func TestAccessTokenMemory(t *testing.T) {
...
+
+			Convey("Finishes without raising errors and data has been deleted", func() {
+				// Write Data to source interface
+				dataActual := []byte{'t', 'e', 's', 't'}
+				tokenInMemory.Write(dataActual)
 
-			// Set proxy struct with source interface
-			var proxyStruct ProxyHandlerTokenMemory
-			proxyStruct.AccessToken = tokenInMemory
+				// Set proxy struct with source interface
```
orenbm: when

pkg/access_token/access_token_memory_test.go
```diff
@@ -16,84 +16,124 @@ func TestAccessTokenMemory(t *testing.T) {
...
+				var proxyStruct ProxyHandlerTokenMemory
+				proxyStruct.AccessToken = tokenInMemory
 
-			// Delete access token from proxy
-			err := proxyStruct.AccessToken.Delete()
-			So(err, ShouldEqual, nil)
+				// Delete access token from proxy
```
orenbm: when

pkg/access_token/access_token_memory_test.go
```diff
@@ -16,84 +16,124 @@ func TestAccessTokenMemory(t *testing.T) {
...
+				err := proxyStruct.AccessToken.Delete()
+				So(err, ShouldEqual, nil)
 
-			// Data in source interface should be deleted
-			dataExpected, err := tokenInMemory.Read()
-			So(err.Error(), ShouldEqual, "error reading access token, reason: data is empty")
-			So(dataExpected, ShouldEqual, nil)
+				// Data in source interface should be deleted
```
orenbm: then

pkg/access_token/access_token_file_test.go
sigalsax: The last convey here doesn't do what you suggested the message should be.

`Writes the access token to a file in the path defined` should be when the actual writing is taking place. Here we are just checking if the file actually exists. What say you?

pkg/authenticator/authenticator_test.go
```diff
@@ -27,26 +27,35 @@ func parseCert(filename string) (*x509.Certificate, error) {
...
 
-			authn := Authenticator{
-				PublicCert: goodCert,
-			}
+			Convey("Finishes without raising an error and returns certificate", func() {
+				So(err, ShouldBeNil)
+			})
 
-			So(authn.IsCertExpired(), ShouldEqual, false)
+			Convey("Returns the false that the certificate is not expired", func() {
```
orenbm: i think that "Returns false" is enough

pkg/authenticator/authenticator_test.go
```diff
@@ -27,26 +27,35 @@ func parseCert(filename string) (*x509.Certificate, error) {
...
 
-			authn := Authenticator{
-				PublicCert: expiredCert,
-			}
+			Convey("Finishes without raising an error and returns certificate", func() {
+				So(err, ShouldBeNil)
+			})
 
-			So(authn.IsCertExpired(), ShouldEqual, true)
+			Convey("Returns true that the certificate is expired", func() {
```
orenbm: Returns true

pkg/access_token/access_token_file_test.go
orenbm: maybe read the content of the file?

pkg/access_token/access_token_file_test.go
sigalsax: I don't think these should be broken up into 2 because the test isn't focused on reading, it's focused on writing... I think instead of checking if Read was done correct it will be:
```
Convey("Writes the data the file as expected", func() {
 	dataExpected, _ := tokenInFile.Read()
	eq := reflect.DeepEqual(dataActual, dataExpected)
	So(eq, ShouldEqual, true)
})
```

and just get rid of the check of `err` because if Read wasn't done correctly, it will fail there

pkg/access_token/access_token_file_test.go
sigalsax: Instead of `Given two instances` `When two instances of` ?

pkg/access_token/access_token_file_test.go
sigalsax: what do you mean here?

pkg/access_token/access_token_file_test.go
orenbm: good point

pkg/access_token/access_token_file_test.go
orenbm: no. i was referring to "Set proxy struct with source interface". you're saying that you're doing something so it can go inside a "when convey"

pkg/access_token/access_token_file_test.go
orenbm: add an "it" convey for this

pkg/access_token/access_token_memory_test.go
sigalsax: we check it in read tests above but it's more of a side test that should be actually really tested in write. Good catch. Adding now.

pkg/access_token/access_token_memory_test.go
sigalsax: when what?

pkg/authenticator/authenticator_test.go
sigalsax: I am aiming for readability. Returning false doesn't really explain that false = not expired

pkg/authenticator/authenticator_test.go
sigalsax: see above

pkg/access_token/access_token_memory_test.go
orenbm: this should be a when clause

pkg/authenticator/authenticator_test.go
orenbm: yes but when you read the whole context you see:
```
Given a non-expired certificate to authenticate with Conjur
    Returns false
```

i think this is better than:
```
Given a non-expired certificate to authenticate with Conjur
    Returns the false that the certificate is not expired
```

also, it is written in bad english :)

orenbm: it is clear that `false == not expired` because you write that this is your expectation

pkg/access_token/access_token_file_test.go
sigalsax: Can't it be "Checks that the file exists"? If it fails we know that the file doesn't exist

orenbm: yes but then we don't check that the `write` operation succeeds

pkg/authenticator/authenticator_test.go
sigalsax: fixed

pkg/access_token/access_token_file_test.go
sigalsax: I do see value in this test
Added the following:
```
Convey("Checks that the file exists in the path defined", func (){
	// Check if file exists
	_, err = os.Stat("/tmp/accessTokenFileWrite1")
	So(err, ShouldEqual, nil)
})

Convey("When running Read method", func() {
	dataExpected, _ := tokenInFile.Read()

	// Confirm data was written
	Convey("Returns the data the was written to the file", func() {
		eq := reflect.DeepEqual(dataActual, dataExpected)
		So(eq, ShouldEqual, true)
	})
})
```

pkg/access_token/access_token_file_test.go
```diff
@@ -19,138 +19,223 @@ func TestAccessTokenFile(t *testing.T) {
...
-		Convey("Returns true if Data output as expected", func() {
+
+		Convey("Given an access token with data", func() {
 			dataActual := []byte{'t', 'e', 's', 't'}
 			err := tokenInFile.Write(dataActual)
-			So(err, ShouldEqual, nil)
 
-			dataExpected, err := tokenInFile.Read()
-			eq := reflect.DeepEqual(dataActual, dataExpected)
+			Convey("Finishes without raising an error", func() {
```
orenbm: this is actually part of the given clause of this test. can we change it to `That finishes without raising an error`?

orenbm: we should make clear what we test in each tet

pkg/access_token/access_token_file_test.go
```diff
@@ -19,138 +19,223 @@ func TestAccessTokenFile(t *testing.T) {
...
-			So(err, ShouldEqual, nil)
+			Convey("When running the Write method", func() {
+				tokenInFile.TokenFilePath = "/tmp/accessTokenFileWrite1"
+				err := tokenInFile.Write(dataActual)
+
+				Convey("Finishes without raising an error", func() {
+					So(err, ShouldEqual, nil)
+				})
+
+				Convey("Checks that the file exists in the path defined", func (){
```
orenbm: the sentence is wrong. a better one will be `Creates a file in the path defined`.

Note that we're not explaining what the test does but what the method that we're testing does

pkg/access_token/access_token_file_test.go
```diff
@@ -19,138 +19,223 @@ func TestAccessTokenFile(t *testing.T) {
...
+					// Check if file exists
```
orenbm: you can remove this comment, it is obvious

pkg/access_token/access_token_file_test.go
```diff
@@ -19,138 +19,223 @@ func TestAccessTokenFile(t *testing.T) {
...
+
+				Convey("When running Read method", func() {
+					dataExpected, _ := tokenInFile.Read()
+
+					// Confirm data was written
+					Convey("Returns the data the was written to the file", func() {
+						eq := reflect.DeepEqual(dataActual, dataExpected)
+						So(eq, ShouldEqual, true)
+					})
+				})
```
orenbm: I'd rather have this as:
```
Convey("Writes the data to the file", func() {
	// TODO: read the content with `os` methods (not with `tokenInFile`)

	eq := reflect.DeepEqual(dataActual, dataExpected)
	So(eq, ShouldEqual, true)
})
```

orenbm: This way it explains what we are testing and we won;'t fail on `tokenInFile.Read()` errors. We are pin-pointing the test to what is tested without noises

pkg/access_token/access_token_file_test.go
```diff
@@ -19,138 +19,223 @@ func TestAccessTokenFile(t *testing.T) {
...
 
-			So(err.Error(), ShouldEqual, "error writing access token, reason: data is empty")
+				Convey("Raises an error that the access token data is empty", func() {
+					So(err.Error(), ShouldEqual, "error writing access token, reason: data is empty")
+				})
+			})
 		})
 
-		Convey("Returns no error if file already exists", func() {
+		Convey("Given access token with data and a file location defined", func() {
```
orenbm: what's the difference between this sceanrio and the one above? Can't we just add the "When running the Write method a second time" convey in the scenario above?

pkg/access_token/access_token_file_test.go
```diff
@@ -19,138 +19,223 @@ func TestAccessTokenFile(t *testing.T) {
...
+					_, err = os.Stat("/tmp/accessTokenFileWrite2")
+					So(err, ShouldEqual, nil)
+				})
+			})
+
+			Convey("When running the Write method a second time", func() {
+				dataActual = []byte{'t', 'e', 's', 't', '2'}
+				err := tokenInFile.Write(dataActual)
+
+				Convey("Creates with file without raising an error", func() {
```
orenbm: Creates the file

pkg/access_token/access_token_file_test.go
```diff
@@ -19,138 +19,223 @@ func TestAccessTokenFile(t *testing.T) {
...
+					// Check if file exists
+					_, err = os.Stat("/tmp/accessTokenFileWrite2")
+					So(err, ShouldEqual, nil)
+				})
+
+				Convey("Writes the data the file as expected", func() {
```
orenbm: to the file

pkg/access_token/access_token_file_test.go
```diff
@@ -19,138 +19,223 @@ func TestAccessTokenFile(t *testing.T) {
...
+					dataExpected, _ := tokenInFile.Read()
```
orenbm: i'd rather do this with `os` read

pkg/access_token/access_token_file_test.go
```diff
@@ -19,138 +19,223 @@ func TestAccessTokenFile(t *testing.T) {
...
+		Convey("Given access token with data", func() {
+			dataActual := []byte{'t', 'e', 's', 't'}
 
-			// Both input & output arrays should be cleared from memory
-			empty := make([]byte, len(dataActual))
-			eq := reflect.DeepEqual(dataActual, empty)
-			So(eq, ShouldEqual, true)
-			eq = reflect.DeepEqual(dataFromRead, empty)
-			So(eq, ShouldEqual, true)
+			Convey("When running the Write method", func() {
```
orenbm: this is given and not when. in this test we're testing the `delete` funcionality. the fact that it got there by running `write` is Given to us.

pkg/access_token/access_token_file_test.go
```diff
@@ -19,138 +19,223 @@ func TestAccessTokenFile(t *testing.T) {
...
+				tokenInFile.TokenFilePath = "/tmp/accessTokenFileDel1"
+				err := tokenInFile.Write(dataActual)
+
+				Convey("Finishes without raising an error", func() {
+					So(err, ShouldEqual, nil)
```
orenbm: this can go inside the convey above.
```
Convey("And the data was written successfully", func() {
	tokenInFile.TokenFilePath = "/tmp/accessTokenFileDel1"
	err := tokenInFile.Write(dataActual)
	So(err, ShouldEqual, nil)
})
```

pkg/access_token/access_token_file_test.go
orenbm: see suggestion below

pkg/access_token/access_token_file_test.go
```diff
@@ -19,138 +19,223 @@ func TestAccessTokenFile(t *testing.T) {
...
+				})
+
+				Convey("When running the Read method", func() {
```
orenbm: why do we need this? we're not testing here the `read` method. if we do need it, it should be Given (but i don't think we need it anyway)

pkg/access_token/access_token_file_test.go
```diff
@@ -19,138 +19,223 @@ func TestAccessTokenFile(t *testing.T) {
...
+					dataFromRead, err := tokenInFile.Read()
+
+					Convey("Finishes without raising an error", func() {
+						So(err, ShouldEqual, nil)
+					})
+
+					Convey("When running the Delete method", func() {
```
orenbm: _this_ is the When :)

pkg/access_token/access_token_file_test.go
```diff
@@ -19,138 +19,223 @@ func TestAccessTokenFile(t *testing.T) {
...
+						err = tokenInFile.Delete()
+
+						Convey("Finishes without raising an error", func() {
+							So(err, ShouldEqual, nil)
+						})
+
+						Convey("Properly deletes the file", func() {
+							_, err = os.Stat("/tmp/accessTokenFileDel1")
+							So(err, ShouldNotEqual, nil)
```
orenbm: do we want to check the actual error? to verify it doesn't fail oin something else and the file is not deleted

pkg/access_token/access_token_file_test.go
```diff
@@ -19,138 +19,223 @@ func TestAccessTokenFile(t *testing.T) {
...
+							eq = reflect.DeepEqual(dataFromRead, empty)
+							So(eq, ShouldEqual, true)
+						})
+					})
+				})
+			})
 		})
 
-		Convey("Returns no error if Data input is nil", func() {
+		Convey("Given an access token with no data saved to a file", func() {
```
orenbm: how does this happen?

pkg/access_token/access_token_file_test.go
```diff
@@ -19,138 +19,223 @@ func TestAccessTokenFile(t *testing.T) {
...
-
-			// Check that file is not exits
-			err = tokenInFile.Delete()
-			So(err.Error(), ShouldEqual, "error deleting access token")
+				Convey("Deletes the file and no errors are raised", func() {
+					So(err, ShouldEqual, nil)
+
+					// Check that file does not exist
+					_, err = os.Stat("/tmp/accessTokenFileDel2")
+					So(err, ShouldNotEqual, nil)
```
orenbm: same here- maybe check the actual error

pkg/access_token/access_token_file_test.go
```diff
@@ -19,138 +19,223 @@ func TestAccessTokenFile(t *testing.T) {
...
+				})
+
+				Convey("When running the Delete method again on the same file", func() {
+					err = tokenInFile.Delete()
+
+					Convey("Finishes with an error deleting the access token", func() {
```
orenbm: i think `Raises an error` is enough

pkg/access_token/access_token_memory_test.go
```diff
@@ -16,84 +16,143 @@ func TestAccessTokenMemory(t *testing.T) {
...
 			dataActual := []byte{'t', 'e', 's', 't'}
-			err := tokenInMemory.Write(dataActual)
 
-			So(err, ShouldEqual, nil)
+			Convey("Writes the access token to memory without raising an error", func() {
+				err := tokenInMemory.Write(dataActual)
+				So(err, ShouldEqual, nil)
+			})
+
+			Convey("When running Read method", func() {
```
orenbm: this should go inside the convey above. also, please see my comment on this in `access_token_file`

orenbm: (read with `os`, not `tokenInFile`

pkg/access_token/access_token_memory_test.go
```diff
@@ -16,84 +16,143 @@ func TestAccessTokenMemory(t *testing.T) {
...
+		Convey("Given an access token with data saved in memory", func() {
+			dataActual := []byte{'t', 'e', 's', 't'}
 
-			// Both input & output arrays should be cleared from memory
-			empty := make([]byte, len(dataActual))
-			eq := reflect.DeepEqual(dataActual, empty)
-			So(eq, ShouldEqual, true)
-			eq = reflect.DeepEqual(dataFromRead, empty)
-			So(eq, ShouldEqual, true)
+			Convey("When running the Write method", func() {
```
orenbm: same - this isn't a when. it is given in this test

pkg/authenticator/authenticator_test.go
```diff
@@ -27,26 +27,35 @@ func parseCert(filename string) (*x509.Certificate, error) {
...
+
+		Convey("Given a non-expired certificate to authenticate with Conjur", func() {
 			goodCert, err := parseCert("testdata/good_cert.crt")
-			So(err, ShouldBeNil)
 
-			authn := Authenticator{
-				PublicCert: goodCert,
-			}
+			Convey("Finishes without raising an error and returns certificate", func() {
+				So(err, ShouldBeNil)
```
orenbm: where do we verify that it returns a certificate?

pkg/authenticator/authenticator_test.go
```diff
@@ -27,26 +27,35 @@ func parseCert(filename string) (*x509.Certificate, error) {
...
 
-			authn := Authenticator{
-				PublicCert: expiredCert,
-			}
+			Convey("Finishes without raising an error and returns certificate", func() {
+				So(err, ShouldBeNil)
+			})
 
-			So(authn.IsCertExpired(), ShouldEqual, true)
+			Convey("Returns true", func() {
```
orenbm: "Returns true on `IsCertExpired`" is better

pkg/authenticator/authenticator_test.go
```diff
@@ -27,26 +27,35 @@ func parseCert(filename string) (*x509.Certificate, error) {
...
 
-			authn := Authenticator{
-				PublicCert: goodCert,
-			}
+			Convey("Finishes without raising an error and returns certificate", func() {
+				So(err, ShouldBeNil)
+			})
 
-			So(authn.IsCertExpired(), ShouldEqual, false)
+			Convey("Returns false", func() {
```
orenbm: "Returns false on `IsCertExpired`" is better

pkg/access_token/access_token_file_test.go
sigalsax: i added in previous commit the read method that confirms that the data was written to the file

pkg/access_token/access_token_file_test.go
sigalsax: I am not sure I understand here. Make it part of the `given` or the 'it' is implicit clause?

pkg/access_token/access_token_file_test.go
sigalsax: but I don't think that is what this line is testing. `os.Stat` retrieves stat info about the file and if there is no error, then we know the file exists. I don't here we are creating the file otherwise it would be `WriteFile()`

pkg/access_token/access_token_file_test.go
sigalsax: I thought TODO is something that is awaiting implementation. If I remove `dataExpected, _ := tokenInFile.Read()` I won't be able to compare if it was the proper result. Suggest putting all  commented out?

pkg/access_token/access_token_file_test.go
sigalsax: joining those tests now and deleting the latter one since they basically the same, just one with an added complexity

pkg/access_token/access_token_file_test.go
sigalsax: why? we want to mock our `read` function not `os.read`

pkg/access_token/access_token_file_test.go
sigalsax: so not `Convey("When running the Write method", func() {...`?

pkg/access_token/access_token_file_test.go
sigalsax: So it is less about what is actually run and more about what it means to run the Write method? In order words `And the data was written successfully` instead of `When running the write method`?

pkg/access_token/access_token_file_test.go
sigalsax: why don't need this in a `convey` or just not all? Nessi wanted to test end-to-end with write, read, delete so that's why we wanted the read method in there but we can probably not worry about the testing if there is an err because that should be taken care of when `Read` is actually tested above

pkg/access_token/access_token_file_test.go
sigalsax: so this is a good thing? haha no changes required? ohhhh i understand, the when is the function we are checking.

pkg/access_token/access_token_file_test.go
sigalsax: adding but what other errors could it be besides `error deleting access token` (that's what we wrote in `Delete`)? Maybe os specific?

pkg/access_token/access_token_file_test.go
orenbm: Given. it's not what you're actually testing here. it is given that it finishes without raising an error *and only then* you start the actual test

pkg/access_token/access_token_file_test.go
orenbm: we're not creating the file. the method that we're testing is creating the file and that's what we're validating here. you've done it right in `access_token_memory_test.go`

pkg/access_token/access_token_file_test.go
orenbm: TODO by you now :)

orenbm: what i meant is that we should read with `os` methods and not with our methods. this is because it's better to assume `os` methods work rather than our methods

pkg/access_token/access_token_file_test.go
sigalsax: which scenario? that an access token is _saved_ without any data or that an access token is nil?

pkg/access_token/access_token_file_test.go
orenbm: yes but the test here is on `write` and the `read` part is given. i'd rather not fall on given stages because of our failures

pkg/access_token/access_token_file_test.go
orenbm: yes, exactly

pkg/access_token/access_token_file_test.go
sigalsax: check the actual os.stat error?

pkg/access_token/access_token_file_test.go
orenbm: yes

pkg/access_token/access_token_file_test.go
orenbm: yes, i would remove `read` from this test

pkg/access_token/access_token_file_test.go
orenbm: yep

pkg/access_token/access_token_file_test.go
sigalsax: deleted previous test so this doesnt apply anymore

pkg/access_token/access_token_file_test.go
orenbm: yes, os specific

pkg/authenticator/authenticator_test.go
sigalsax: that is what parseCert returns, a cert and an error.

pkg/authenticator/authenticator_test.go
sigalsax: I like this change

PR #40, opened 2019-08-28
Extract access token handling & read from env

orenbm:
This PR includes:
- Read tokenFilePath & clientCertPath as env vars
- Extract access token handling to its own package


cmd/authenticator/main.go
```diff
@@ -18,20 +18,20 @@ var infoLogger = authenticator.InfoLogger
 func main() {
 	var err error
 
-	// Parse any flags for client cert / token paths, and set default values if not passed
-	clientCertPath := flag.String("c", "/etc/conjur/ssl/client.pem",
```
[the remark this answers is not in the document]
orenbm: yes, ultimately. not only these vars but all of them. i'd rather do that properly with a thourough design because it raises some question (i.e which input is higher in priority if both inputs are given).

i think it's better to have all the inputs as env vars and then re-think the whole thing. here is the ticket: https://github.com/cyberark/conjur/issues/1142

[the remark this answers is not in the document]
orenbm: I'd rather not have half-baked features in master (adds complexity and requires more tests). if these parameters are not used as cmd inputs at the moment i'd rather fully design this and re-add them later with all the rest.
@izgeri @sgnn7 @jonahx @doodlesbykumbi do you know of any usage of `clientCertPath` & `tokenFilePath` via args or do we always use the default value?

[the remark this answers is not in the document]
doodlesbykumbi: @orenbm Given that the image for this client is out in the world at large, perhaps there are users leveraging the flags already. I'm not aware of any usage of the flags on out side.

pkg/authenticator/authenticator.go
```diff
@@ -9,6 +9,7 @@ import (
 	"encoding/asn1"
 	"encoding/pem"
 	"fmt"
+	"github.com/cyberark/conjur-authn-k8s-client/pkg/access_token"
```
doodlesbykumbi: Some linting comments:

1. Splitting imports to "go standard", "external deps" and "internal" improves readability of dependencies
```
import (
  ## go standard
  "encoding/pem"

  # external
  "gopkg.in/yaml.v2"

  # internal
  "github.com/cyberark/conjur-authn-k8s-client/pkg/access_token"
)
```

2. Similar to `#1`, sort (alphabetically) and categorise aggressively (imports, struct field names etc.). Below the struct definition is split into private fields and public fields, sorted alphabetically.
```
type Authenticator struct {
  client             *http.Client
  privateKey         *rsa.PrivateKey
  AccessTokenHandler access_token.AccessTokenHandler
  Config             authnConfig.Config
  PublicCert         *x509.Certificate
}
```

pkg/access_token/access_token_file_test.go
```diff
@@ -0,0 +1,237 @@
...
+	"os"
+	"reflect"
+	"testing"
+)
+
+type ProxyHandlerTokenFile struct {
+	AccessToken AccessTokenHandler
+}
+
+func TestAccessTokenFile(t *testing.T) {
```
doodlesbykumbi: There seems to be enough overlap between the tests for AccessTokenFile and AccessTokenMemory to warrant splitting out tests for any given `AccessTokenHandler` implementation to
1. Tests that validate the behavior expected of all AccessTokenHandlers
2. Tests that validate the behavior that is specific to this particular implementation e.g. Checking that a file is indeed delete for the AccessTokenFile

That should DRY things up a bit.

pkg/access_token/access_token_handler.go
```diff
@@ -0,0 +1,7 @@
+package access_token
+
+type AccessTokenHandler interface {
```
doodlesbykumbi: 1. It's not clear to me why AccessTokenHandler does from its name. It would be useful to have a brief description of AccessTokenHandler for godocs.
2. Reading this PR it's unclear why Read and Delete are necessary. I haven't seen them be exercised outside of the unit tests for a given AccessTokenHandler implementation.

pkg/authenticator/authenticator.go
```diff
@@ -236,8 +247,7 @@ func (auth *Authenticator) ParseAuthenticationResponse(response []byte) error {
 		content = response
 	}
 
-	// InfoLogger.Printf("Writing token %v to shared volume ...", content)
-	err = ioutil.WriteFile(auth.Config.TokenFilePath, content, 0644)
+	err = auth.AccessTokenHandler.Write(content)
```
doodlesbykumbi: I know this is out of scope of this PR. 
The method in which this call takes place has a misleading name (ParseAuthenticationResponse). It's definitely not merely parsing the authentication response.

pkg/access_token/access_token_memory.go
```diff
@@ -0,0 +1,41 @@
...
+	Data []byte
+}
+
+func NewAccessTokenMemory() (token *AccessTokenMemory, err error) {
+	return &AccessTokenMemory{
+		Data: nil,
+	}, nil
+}
+
+func (token AccessTokenMemory) Read() (Data []byte, err error) {
```
doodlesbykumbi: This is happening in a few places. The named returns aren't being used so they might as-well not be named. Unused named returns
1. Unnecessarily assign variables (I think but maybe the compiler just optimises them out) 
2. Can be misleading since the reader might be expecting to follow those named returns to understand what gets returned

pkg/access_token/access_token_file.go
```diff
@@ -0,0 +1,69 @@
...
+
+	token.Data = Data
+	// Write the data to file
+	// Create the directory if it doesn't exist
+	tokenDir := filepath.Dir(token.TokenFilePath)
+	if _, err := os.Stat(tokenDir); os.IsNotExist(err) {
+		err = os.MkdirAll(tokenDir, 755)
+		if err != nil {
+			// Do not specify the directory in the error message for security reasons
+			return fmt.Errorf("error writing access token, reason: failed to create directory")
```
doodlesbykumbi: There's a few places where the error returned is an interpretation of the original error. I think that might have the effect of making debugging more difficult. I think it might warranted to return an interpreted error in certain circumstances. I might be wrong but I'm not immediately convinced that file paths and folder paths warrant it.

pkg/access_token/access_token_file_test.go
orenbm: not sure i understand. what will be the files that you'll have? can you write here some template?

pkg/access_token/access_token_handler.go
orenbm: Well, actually they are not used here. they are used in the project that will consume this authenticator. The interface needs to be here so objects can be created in the consuming project. 

For example, in the `conjur-k8s-secrets-manager` we will use `AccessTokenMemory`. we will create the `authenticator` object with this handler which will not write the data to a file. later on, we will use the `Read()` method to get the token for retrieving secrets from conjur.

developers who consume this project can create their own `AccessTokenHandler` but these two are used by our projects so they are fully integrated in the project. think about it like with ansible. you can create your own plugin and if it's good enough it will be implemented inside the ansible project.

orenbm: > It's not clear to me why AccessTokenHandler does from its name.

isn;t it clear from its methods? not sure what i can write besides 
```
This interface handles a Conjur access token. Classes implementing this interface should have the ability to read, write & delete the access token
```

pkg/authenticator/authenticator.go
orenbm: it's very risky to change this now as it's used in external projects

pkg/access_token/access_token_file.go
orenbm: not sure i got your comment here. what's the action item?

doodlesbykumbi: What is the security argument for obscuring the original error with folder path and file name ?

I was thinking we should send the interpreted error concatenated with the original error. The errors coming from deeper in the stack provide meaningful information for debugging.

pkg/authenticator/authenticator.go
doodlesbykumbi: That's fair and actually raises an important consideration. External projects using this client really ought to consume an interface so that we can work on the internals free from introducing breaking changes.

pkg/access_token/access_token_handler.go
doodlesbykumbi: Adding examples, as you did in your first response comment, makes it all the more concrete. It's this sort of description which would be useful for someone to grok the utility of the interface - the rich context that comes from examples is invaluable.

pkg/access_token/access_token_file_test.go
doodlesbykumbi: Something like below. The idea being to leverage a [table driven tests](https://github.com/golang/go/wiki/TableDrivenTests ) like pattern 
```
type AccessTokenHandler interface {}
type AccessTokenFile struct {}
type AccessTokenMemory struct {}

type AccessTokenHandlerTestCase struct {
	name  string
	handlerConstructor func() AccessTokenHandler
}
var accessTokenHandlerTests = []AccessTokenHandlerTestCase{
	{
		"AccessTokenMemory",
		func() AccessTokenHandler {
			return AccessTokenMemory{}
		},
	},
	{
		"AccessTokenMemory",
		func() AccessTokenHandler {
			return AccessTokenFile{}
		},
	},
}

func TestAccessTokenImplentation(t *testing.T, tc AccessTokenHandlerTestCase) {
	Convey("Read", t, func() {
		accessTokenHandler := tc.handlerConstructor()

			Convey("Given an access token with data saved in memory", func() {
			dataActual := []byte{'t', 'e', 's', 't'}
			_ = accessTokenHandler.Write(dataActual)

			Convey("When running Read method", func() {
				dataExpected, err := accessTokenHandler.Read()

				Convey("Finishes without raising an error", func() {
					So(err, ShouldEqual, nil)
				})

				Convey("Returns the data the was written", func() {
					eq := reflect.DeepEqual(dataActual, dataExpected)
					So(eq, ShouldEqual, true)
				})
			})
		})

	})
}
func TestAccessTokenImplentations(t *testing.T) {
	for _, at := range accessTokenHandlerTests {
		TestAccessTokenImplentation(t, at)
	}
}
```

pkg/access_token/access_token_handler.go
orenbm: cool, i added a description

cmd/authenticator/main.go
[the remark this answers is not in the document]
orenbm: ok i re-added this ability. thanks!

pkg/authenticator/authenticator.go
orenbm: done

pkg/access_token/access_token_file_test.go
orenbm: i actually think this decreases readability. the only method that is the same (for now, future implementations may differ) is `Read` and the rest will be in its own file anyway. i think that having each class's test in its own file is better also for readability and for future development

pkg/access_token/access_token_file.go
orenbm: @shaharglazner can you please take a look and see if we have some security concern here?

pkg/access_token/access_token_handler.go
```diff
@@ -0,0 +1,17 @@
+package access_token
+
+/*
+	This interface handles a Conjur access token. Classes implementing this interface should have the ability to read,
```
doodlesbykumbi:
```suggestion
	This interface handles a Conjur access token. Structs implementing this interface should have the ability to read,
```

pkg/access_token/access_token_handler.go
doodlesbykumbi: @orenbm Thanks for including the description. I was thinking we could provide a description that alludes to example usage of the methods e.g. `conjur-k8s-secrets-manager` or Secretless. When I initially thought about the example I was looking towards this:

> For example, in the conjur-k8s-secrets-manager we will use AccessTokenMemory. we will create the authenticator object with this handler which will not write the data to a file. later on, we will use the Read() method to get the token for retrieving secrets from conjur.

pkg/access_token/access_token_file_test.go
doodlesbykumbi: I'm starting to agree with you here. Looking at Write and Delete those seem to have very implementation specific. 

My reasoning was that implementations of an interface implicitly have an expected behavior when the methods are called wherever they are called. There would be tests that would validate this. These tests would be run on any new implementation as a smoke test/sanity check - table driven test would allow these to be run in a DRY fashion.

pkg/access_token/access_token_handler.go
izgeri: I'm curious why we don't just have an object `accessToken` that has methods on it `read`, `write`, `delete`. what is the value of separating this out into a "handler"? What is the purpose of having a separate "memory" object?

pkg/access_token/access_token_file_test.go
orenbm: i see but we will still need to have the separate tests. for example, we could have a generic test for `Write` in which we will verify that it wrote the content but then again we will need a separate test for `AccessTokenFile` to verify that the data was written to the file specified. This test will probably be very similar if not identical to the generic one so i don't see a benefit here.

@doodlesbykumbi can you write you action item here? do you still think i should add the generic test file?

pkg/access_token/access_token_handler.go
orenbm: @izgeri just to make sure i understand you, are you suggesting that we have an `AccessToken` object which will hold the logic of wether we should write the token to a file or to the memory? This means that the `AccessToken` needs to be aware of the use-case we're at and will have some `if` clause or switch case in it.

If that's what you mean then i think it is better to have different structs that implement the `handler` interface and have the logic outside. the word `handler` is just for readability as i think it better explains what the object does (it handles an access token it's not an actual access token). 

what do you think?

pkg/access_token/access_token_file.go
shaharglazner: So, there are two security concerns here. I agree these aren't fundamental security concerns, but it worth discussing it.

**The path to the token is somehow sensitive** 
OWASP #8.1 - 
```
Verify that the application does not output error messages or stack traces containing sensitive data that could assist an attacker, including session-id, software/framework versions, and personal information
```
So, the question here is if the path to the token can help the attacker to get it, or it doesn't matter because the attacker would find it anyway. I think it is sensitive in some way, therefore I think its worth to consider to specify the path only in DEBUG or something similar (or to add the ability to remove it from logs)


**The content of the path is a user-input and therefore could be used to manipulate the logs/the system that reads them**
OWASP #8.5 and #8.8 - 
```
8.5
Verify that all events that include untrusted data will not execute as code in the intended log viewing software.
```
```
8.8
Verify that all non-printable symbols and field separators are properly encoded in log entries, to prevent log injection.
```
So if we are going to log the path, we need to validate/sanitize if before just printing it.

WDYT? @orenbm @doodlesbykumbi

pkg/access_token/access_token_handler.go
jonahx: Consider the following a fresh comment: 

My big comment is that the interface currently called `AccessTokenHandler` is actually just defining the concept of an access token.  It should be called `AccessToken`.  Then file and memory are two specific implementation of an access token (re: the conversation above with Geri, I am agnostic as to whether or not having those to implementations is a good idea, that all depends on what you're doing with them).

But assuming that it is a good idea, the package design convention you should follow is to have a structure like this:
```
access_token.go (defines the interface only)
    file
        access_token.go (file impl in file package.  client code, eg, creates one like  file.NewAccessToken)
    memory
        access_token.go (memory imple.  client code does memory.NewAccessTokent(...), etc)
```

So, 

1. root `accesstoken` package defines the interface only and has no dependencies.
2. packaces below that `file` and `memory` define the specific implementations and do have deps.
3. Note that this cleans up the naming as client code does `file.NewAccessToken` instead of `NewAccessTokenFile`.

orenbm: nice!

orenbm: done refactor

pkg/access_token/access_token_file.go
orenbm: i think that the value we can get from printing this is not high enough to take the risk or even further investigate it. I'll wait for @doodlesbykumbi 's thought on this before we proceed.

pkg/access_token/access_token_handler.go
jonahx: Noice!!  You don't mess around.

One small leftover thing:

```
type AccessTokenFile struct {
	Data          []byte
	TokenFilePath string
}
```
can now become `AccessToken` because you're already in the `file` package now.  This is called "package name stutter" and is a common mistake to be avoided.

I might consider changing `TokenFilePath` to `PathToToken` but that's more of a subjective personal suggestion: take it or leave it.

orenbm: done, thanks

jonahx: I spent some time looking at this today, and I have other notes about the code base I'll try to post later, but re: the `file.AccessToken` specifically, I think a couple things struck me as problematic.

1. Afaict, you must call `Write()` before you call `Read()`.  So clients of this object cannot simply create an object and then use it.  They must create the object, _and then_ call `Write()` before the object is really "ready".  Thus the construction responsibility has been split up across two methods, rather than residing in the constructor.  A fun old-school term for this kind of coupling is [Connascence of Execution](https://connascence.io/execution.html).
2. When you `Read()` back the value, you read it from memory.  Indeed, it errors if the token is not in memory:

   ```go
   if token.Data == nil {
       return nil, fmt.Errorf("error reading access token, reason: data is empty")
   }
   ```
   So really, this `file.AccessToken` appears to be fundamentally the same as the `memory.AccessToken`, except when you `Write()` it has the _side effect of creating a file_.  However, I wasn't able to find in the code where this file is ever read after being created.  I may have missed it -- if so, where does this happen?   In any case, the reason for this design I assume is so that in memory data and the file can travel through the program as a single unit, and both will be cleaned up together when `Delete()` is called.  However, I think I'd prefer a different higher-level code structure that made the use the file clearer.  But that would probably require larger changes the code, and I'd want to understand it better before suggesting anything.

nessiLahav: @jonahx hi, thanks for the detailed comment :)
1. You right, there is a  construction responsibility. 
2. The file is never read after bean created, but it should not deleted from the file system, when we are in SECRET_DESTINATION=none mode, for the application to use it to fetch secrets.

i would like to here how do you suggest to redesign it  ?

pkg/authenticator/authenticator.go
```diff
@@ -9,6 +9,7 @@ import (
 	"encoding/asn1"
 	"encoding/pem"
 	"fmt"
+	"github.com/cyberark/conjur-authn-k8s-client/pkg/access_token/file"
```
doodlesbykumbi: please move this down

pkg/access_token/access_token.go
```diff
@@ -0,0 +1,15 @@
+package access_token
+
+/*
+	This interface handles a Conjur access token. Structs implementing this interface should have the ability to read,
+	write & delete the access tokenץ
+
+	For example, in the conjur-k8s-secrets-manager we will use AccessTokenMemory. we will create the authenticator
+	object with this handler which will not write the data to a file. later on, we will use the Read() method to get the
```
doodlesbykumbi:
```suggestion
	object with this handler which will not write the data to a file. Later on, we will use the Read() method to get the
```

pkg/access_token/access_token_file_test.go
doodlesbykumbi: I think it's fine to let it be here.

pkg/access_token/access_token_file.go
doodlesbykumbi: Fair assessment @orenbm. Thanks for giving this some consideration.

pkg/authenticator/authenticator.go
orenbm: nice catch

orenbm:
> Looking good. Left 2 comments. I think after that we should be good to approve.

Done, thanks!

so now we're just waiting for @jonahx 's approval on this conversation: https://github.com/cyberark/conjur-authn-k8s-client/pull/40#discussion_r319846033

pkg/access_token/access_token_handler.go
jonahx: > when we are in SECRET_DESTINATION=none mode, for the application to use it to fetch secrets.

pkg/authenticator/authenticator.go
```diff
@@ -236,27 +248,25 @@ func (auth *Authenticator) ParseAuthenticationResponse(response []byte) error {
 		content = response
 	}
 
-	// InfoLogger.Printf("Writing token %v to shared volume ...", content)
-	err = ioutil.WriteFile(auth.Config.TokenFilePath, content, 0644)
+	err = auth.AccessTokenHandler.Write(content)
```
doodlesbykumbi: Can we rename the field `AccessTokenHandler` on `auth` to `AccessToken` ? To be consistent.

pkg/access_token/access_token.go
```diff
@@ -0,0 +1,15 @@
+package access_token
+
+/*
+	This interface handles a Conjur access token. Structs implementing this interface should have the ability to read,
```
doodlesbykumbi: Let me know what you think of this rephrasing. The intention was to:
1. highlight that AccessToken represents an access token.
2. elaborate on the example a bit
3. strip the first person plural to shift the focus of actor to conjur-k8s-secrets-manager


```
This interface represents a Conjur access token residing in some arbitrary storage medium (e.g. in-app memory or file system). Structs implementing this interface should have the ability to read,
write & delete the access token contents.
For example, conjur-k8s-secrets-manager uses memory.AccessToken. The authenticator
object is created with an empty access token which is populated with the Write method when access token contents become available. In this case the Write method writes the data to in-app memory. Later, conjur-k8s-secrets-manager can call the Read() method to access the
token-contents for retrieving secrets from Conjur.
```

pkg/authenticator/authenticator.go
orenbm: nice catch. i thought i already did this...

PR #47, opened 2019-09-03
Pr comments

orenbm:
Implemented comments from the PR

@nessiLahav, @moticless & @sigalsax - please go over each commit to see the changes.

PR #48, opened 2019-09-03
WIP: Escape on backslashes

sigalsax:
This PR adds logic to escape backslashes before being patched
That way, when a secret contains \, our implementation will not fail
- [x] \ are escaped
- [x] UTs have been added

NOTE: do not MERGE! We are moving repos for CA and these changes will be cherry-picked over to the new repo

### DIFF
--- a/pkg/secrets/k8s/k8s_secrets.go
+++ b/pkg/secrets/k8s/k8s_secrets.go
@@ -3,16 +3,18 @@ package k8s
 import (
 	"bytes"
 	"fmt"
+	"regexp"
+	"strings"
+
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+	"k8s.io/apimachinery/pkg/types"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
 	log "github.com/cyberark/conjur-authn-k8s-client/pkg/logging"
 	secretsConfig "github.com/cyberark/conjur-authn-k8s-client/pkg/secrets/config"
 	"github.com/cyberark/conjur-authn-k8s-client/pkg/utils"
-	v1 "k8s.io/api/core/v1"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/types"
-	"k8s.io/client-go/kubernetes"
-	"k8s.io/client-go/rest"
-	"regexp"
-	"strings"
 )
 
 type K8sSecretsHandler struct {
@@ -186,12 +188,12 @@ func generateStringDataEntry(stringDataEntriesMap map[string][]byte) ([]byte, er
 	entries := make([][]byte, len(stringDataEntriesMap))
 	// Parse every key-value pair in the map to a "key:value" byte array
 	for key, value := range stringDataEntriesMap {
-		stringDataEntriesMap[key] = bytes.Replace(value, []byte("\\"), []byte("\\\\"), -1)
+		value = escapeDataEntry(value)
 		entry = utils.ByteSlicePrintf(
 			`"%v":"%v"`,
 			"%v",
 			[]byte(key),
-			stringDataEntriesMap[key],
+			value,
 		)
 		entries[index] = entry
 		index++
@@ -213,6 +215,7 @@ func generateStringDataEntry(stringDataEntriesMap map[string][]byte) ([]byte, er
 				entries[i+1],
 			)
 		}
+
 		// Clear secret from memory
 		entries[i] = nil
 	}
@@ -227,4 +230,8 @@ func generateStringDataEntry(stringDataEntriesMap map[string][]byte) ([]byte, er
 	// Clear secret from memory
 	entriesCombined = nil
 	return stringDataEntry, nil
+}
+
+func escapeDataEntry(value []byte) ([]byte) {
+	return bytes.ReplaceAll(value, []byte("\\"), []byte("\\\\"))
 }
--- a/pkg/secrets/k8s/k8s_secrets.go
+++ b/pkg/secrets/k8s/k8s_secrets.go
@@ -188,7 +188,7 @@ func generateStringDataEntry(stringDataEntriesMap map[string][]byte) ([]byte, er
 	entries := make([][]byte, len(stringDataEntriesMap))
 	// Parse every key-value pair in the map to a "key:value" byte array
 	for key, value := range stringDataEntriesMap {
-		value = escapeDataEntry(value)
+		value = escapedSecret(value)
 		entry = utils.ByteSlicePrintf(
 			`"%v":"%v"`,
 			"%v",
@@ -232,6 +232,8 @@ func generateStringDataEntry(stringDataEntriesMap map[string][]byte) ([]byte, er
 	return stringDataEntry, nil
 }
 
-func escapeDataEntry(value []byte) ([]byte) {
-	return bytes.ReplaceAll(value, []byte("\\"), []byte("\\\\"))
+// Escape secrets with backslashes
+// Otherwise, patching K8s secrets will fail because backslashes in Conjur secret are not escaped
+func escapedSecret(secretByte []byte) ([]byte) {
+	return bytes.ReplaceAll(secretByte, []byte("\\"), []byte("\\\\"))
 }
--- a/pkg/secrets/k8s/k8s_secrets.go
+++ b/pkg/secrets/k8s/k8s_secrets.go
@@ -1,6 +1,7 @@
 package k8s
 
 import (
+	"bytes"
 	"fmt"
 	log "github.com/cyberark/conjur-authn-k8s-client/pkg/logging"
 	secretsConfig "github.com/cyberark/conjur-authn-k8s-client/pkg/secrets/config"
@@ -185,11 +186,12 @@ func generateStringDataEntry(stringDataEntriesMap map[string][]byte) ([]byte, er
 	entries := make([][]byte, len(stringDataEntriesMap))
 	// Parse every key-value pair in the map to a "key:value" byte array
 	for key, value := range stringDataEntriesMap {
+		stringDataEntriesMap[key] = bytes.Replace(value, []byte("\\"), []byte("\\\\"), -1)
 		entry = utils.ByteSlicePrintf(
 			`"%v":"%v"`,
 			"%v",
 			[]byte(key),
-			value,
+			stringDataEntriesMap[key],
 		)
 		entries[index] = entry
 		index++
@@ -211,7 +213,6 @@ func generateStringDataEntry(stringDataEntriesMap map[string][]byte) ([]byte, er
 				entries[i+1],
 			)
 		}
-
 		// Clear secret from memory
 		entries[i] = nil
 	}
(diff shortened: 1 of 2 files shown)


pkg/secrets/k8s/k8s_secrets_test.go
```diff
@@ -1,13 +1,13 @@
 package k8s
 
 import (
-	. "github.com/smartystreets/goconvey/convey"
 	log "github.com/cyberark/conjur-authn-k8s-client/pkg/logging"
+	. "github.com/smartystreets/goconvey/convey"
```
orenbm: this is something i'm doing in my PR (https://github.com/cyberark/conjur-authn-k8s-client/pull/47) but in case yours gets merged after mine - please fix `imports` to be sorted in the following way:
- Split into categories (go stdlib => external => internal)
- Order each category alphabetically

see this commit for an example: https://github.com/cyberark/conjur-authn-k8s-client/pull/47/commits/ef46d83efe120cf5d4f537c0f185427e5a242d61

pkg/secrets/k8s/k8s_secrets_test.go
```diff
@@ -41,6 +41,26 @@ func TestKubernetesSecrets(t *testing.T) {
 			})
 		})
 
+		Convey("Given a map of data entries and a secret with backslashes", func() {
+			m := make(map[string][]byte)
```
orenbm: can we give `m` a proper name? what does it mock? i think we have this in several places so please fix that too

pkg/secrets/k8s/k8s_secrets_test.go
```diff
@@ -41,6 +41,26 @@ func TestKubernetesSecrets(t *testing.T) {
...
+			// need to simulate a conjur secret, with 1 backslash, so we use unicode to simulate this
+			m["user"] = []byte("super\u005csecret");
+			DataEntry, err := generateStringDataEntry(m);
```
orenbm: this can be `dataEntry`.

orenbm: it's not exposed

pkg/secrets/k8s/k8s_secrets_test.go
```diff
@@ -41,6 +41,26 @@ func TestKubernetesSecrets(t *testing.T) {
...
+
+			Convey("Finishes without raising an error", func() {
+				So(err, ShouldEqual, nil)
+			})
+
+			Convey("Returns proper password with escaped characters", func() {
+				// before sending secret to k8s there are two backslashes
```
orenbm: secrets? a secret?

pkg/secrets/k8s/k8s_secrets.go
```diff
@@ -1,6 +1,7 @@
 package k8s
 
 import (
+	"bytes"
```
orenbm: please sort this according to this commit (https://github.com/cyberark/conjur-authn-k8s-client/pull/47/commits/ef46d83efe120cf5d4f537c0f185427e5a242d61)

pkg/secrets/k8s/k8s_secrets.go
```diff
@@ -185,11 +186,12 @@ func generateStringDataEntry(stringDataEntriesMap map[string][]byte) ([]byte, er
 	entries := make([][]byte, len(stringDataEntriesMap))
 	// Parse every key-value pair in the map to a "key:value" byte array
 	for key, value := range stringDataEntriesMap {
+		stringDataEntriesMap[key] = bytes.Replace(value, []byte("\\"), []byte("\\\\"), -1)
```
orenbm: please move this to a separate method and add an explanation there on why we are doing this

orenbm: also, i think that running `bytes.Replace` with -1 is like running `bytes.ReplaceAll`. Am I wrong?

orenbm: i guess this works if the UT is passing but can you explain why we are replacing `\\` with `\\\\` and not `\` with `\\`? seems to me that an input with just 1 backslash will stay that way.

orenbm: why do you update `stringDataEntriesMap[key]` instead of editing `value`? is this map entry used later in the code?

pkg/secrets/k8s/k8s_secrets.go
```diff
@@ -211,7 +213,6 @@ func generateStringDataEntry(stringDataEntriesMap map[string][]byte) ([]byte, er
 				entries[i+1],
 			)
 		}
-
```
orenbm: can you please leave this line? improves readability a bit.

pkg/secrets/k8s/k8s_secrets_test.go
```diff
@@ -41,6 +41,26 @@ func TestKubernetesSecrets(t *testing.T) {
 			})
 		})
 
+		Convey("Given a map of data entries and a secret with backslashes", func() {
+			m := make(map[string][]byte)
+			// need to simulate a conjur secret, with 1 backslash, so we use unicode to simulate this
```
orenbm: why do we need unicode to simulate this? can't we have `[]byte("super\secret");`?

pkg/secrets/k8s/k8s_secrets_test.go
orenbm: improve your english in this paragraph please :)

sigalsax: in terms of english, it is correct. It is just in the passive voice. I will update this to be active

pkg/secrets/k8s/k8s_secrets_test.go
sigalsax: fixed in this file and in codebase

pkg/secrets/k8s/k8s_secrets_test.go
```diff
@@ -41,6 +41,26 @@ func TestKubernetesSecrets(t *testing.T) {
...
+			m["user"] = []byte("super\u005csecret");
```
moticless: Maybe we can also test something like:    "\\\\\\\\a\u005\\\\b\\\c\\\\\"  
To verify algorithm doesn't fail on "corner cases"

pkg/secrets/k8s/k8s_secrets.go
sigalsax: A couple of things:
1. It is a one line of code so is it worth throwing it in another method?
2. I didn't have the proper of go version locally (need 1.12 to use replaceAll). Updated the go version on my end and substituted replace -> replaceAll as noted
3. Because go does not escape a single '\', even if in a string and therefore leads to a error "unknown escape character"
4. Didn't want to touch the index value but will change it now

orenbm: 1. yes, for the explanation. It improves readability. 
2. thanks. Nice catch with the go version!
3. ok
4. thanks

pkg/secrets/k8s/k8s_secrets_test.go
sigalsax: explaining in comments now but essentially
We need to simulate a conjur secret with 1 backslash (*\1234). When typing just  []byte("*\1234"), we get an escaping error. If we send a secret with 2 backslashes, the secret will be sent with 2 backslashes which is not presentative of the conjur secret (*\1234)
therefore a trick is to use unicode escape sequence to send the secret with just 1 backslash

pkg/secrets/k8s/k8s_secrets_test.go
sigalsax: just ran it now and we get: \\a\u005\b\c\
which is a good sign! adding this test now (:

sigalsax:
NOTE: do not merge! We are moving repos for CA and these changes will be cherry-picked over to the new repo

pkg/secrets/k8s/k8s_secrets.go
```diff
@@ -225,6 +229,9 @@ func generateStringDataEntry(stringDataEntriesMap map[string][]byte) ([]byte, er
 
 	// Clear secret from memory
 	entriesCombined = nil
-
 	return stringDataEntry, nil
 }
+
```
orenbm: add a description. why do we need this?

pkg/secrets/k8s/k8s_secrets.go
```diff
@@ -225,6 +229,9 @@ func generateStringDataEntry(stringDataEntriesMap map[string][]byte) ([]byte, er
...
+func escapeDataEntry(value []byte) ([]byte) {
+	return bytes.ReplaceAll(value, []byte("\\"), []byte("\\\\"))
+}
```
orenbm: add new line

pkg/secrets/k8s/k8s_secrets.go
orenbm: also, the name `value` is too generic. what are we parsing here? please re-think of the word `escapeDataEntry` too

sigalsax: `value` -> `secretByte`
`escapedDataEntry` -> `escapeSecret`

pkg/secrets/k8s/k8s_secrets_test.go
```diff
@@ -41,11 +42,52 @@ func TestKubernetesSecrets(t *testing.T) {
...
+				stringDataEntryActual := string(escapedDataEntry)
+				eq := reflect.DeepEqual(stringDataEntryActual, stringDataEntryExpected)
+				So(eq, ShouldEqual, true)
+			})
+		})
+
+		// a more complex use-case where the secret contains a combination of backslashes and the unicode escape sequence for backslashes
+		Convey("Given a map of data entries and a secret with unicode and backslashes", func() {
+			dataEntries := make(map[string][]byte)
+			// simulates a Conjur secret with 1 backslash with unicode
```
nessiLahav: incorrect comment

pkg/secrets/k8s/k8s_secrets_test.go
```diff
@@ -41,11 +42,52 @@ func TestKubernetesSecrets(t *testing.T) {
...
+			escapedDataEntry, err := generateStringDataEntry(dataEntries);
+
+			Convey("Finishes without raising an error", func() {
+				So(err, ShouldEqual, nil)
+			})
+
+			Convey("Returns proper password with escaped characters", func() {
+				// before sending a secret to K8s, each secret will be escaped with an additional backslash
+				// K8s trims the second backslash so in order to mimic this, we are expecting two backslashes
+				stringDataEntryExpected := `{"stringData":{"user":"\\\\a\\u005c\\b\\c\\"}}`
```
nessiLahav: please count the number of backslashes in the actual and in the expected output and compare them, for the test to be more clear, because it is hard to identify the number of the backslashes in the expected string

sigalsax:
new pr: https://github.com/cyberark/cyberark-secrets-provider-for-k8s/pull/4 for new repo

pkg/secrets/k8s/k8s_secrets_test.go
sigalsax: changes can be seen in new pr in new repo: https://github.com/cyberark/cyberark-secrets-provider-for-k8s/pull/4

pkg/secrets/k8s/k8s_secrets_test.go
sigalsax: changes can be seen in new pr in new repo: cyberark/cyberark-secrets-provider-for-k8s#4

PR #51, opened 2019-09-04
Prepare for releasing version 14

orenbm:
- Updated version
- Updated Changelog
- Updated go modules


go.sum
```diff
@@ -1,13 +1,35 @@
 github.com/cenkalti/backoff v2.0.0+incompatible h1:5IIPUHhlnUZbcHQsQou5k1Tn58nJkeJL9U+ig5CHJbY=
```
sgnn7: This file contains additions but no deletions even though some modules have been removed. Also, the build is failing most likely due to some dependency not being properly defined in this file that a docker instance is trying to append.

orenbm: re-ran `go get` to update the files

PR #59, opened 2019-12-01
Send full host-id in CSR request

orenbm:
Connected to cyberark/conjur#1239

In order to authenticate hosts from every policy branch, we need the
authn-client to send the full host-id to the server, and in turn the server builds the conjur host-id
from the host-id given by the client instead of using the prefix {@account}:host:conjur/authn-k8s/#{@service_name}/apps.

We send the suffix in the common-name as we did before and we send the prefix
in the request's header.


pkg/authenticator/authenticator.go
```diff
@@ -119,12 +113,21 @@ func (auth *Authenticator) Login() error {
 
 	log.InfoLogger.Printf(log.CAKC007I, auth.Config.Username)
 
-	csrRawBytes, err := auth.GenerateCSR()
+	// We send the hostname suffix in the CSR's common-name so we maintain backwards compatibility
```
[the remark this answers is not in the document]
orenbm: so it seems that the username of a host is `host/<host-id>`. I'll change all occurrences to `username` as here we have the `host/` prefix.

orenbm:
@jvanderhoof @jonahx please re-review.

I added UTs and comments for visibility.

I didn't write this as a comment (didn't see it fit) but i wrote about it in the commit:
- Integration tests for this change are located in:
  - Conjur server: We deploy Conjur in GKE and send `/inject_client_cert`
                   and `/authenticate` requests in different permutations.
  - kubernetes-conjur-demo: We deploy a Conjur cluster and an `authn-client`
                            pod to see hosts can authenticate with it.

orenbm:
@jvanderhoof can this PR get merged? I addressed all of @jonahx's requests.

pkg/authenticator/authenticator.go
```diff
@@ -119,12 +112,23 @@ func (auth *Authenticator) Login() error {
 
 	log.InfoLogger.Printf(log.CAKC007I, auth.Config.Username)
 
-	csrRawBytes, err := auth.GenerateCSR()
+	// We want to let hosts authenticate with Conjur from anywhere in the policy
```
Tovli: Seems this comment should be in the ticket (which number should be part of this PR name)

pkg/authenticator/authenticator.go
```diff
@@ -119,12 +112,23 @@ func (auth *Authenticator) Login() error {
...
+	// tree. For this to work, we send the full username in the CSR request, instead
+	// of just the username suffix (machine identity) like it was before.
+	// To maintain backwards compatibility, the CSR's common-name still contains
+	// only the username suffix (the machine identity) and the prefix is sent in
+	// a header named "Host-Id-Prefix". This header content will be extracted in
+	// the server and will be appended to the suffix from the CSR's common-name.
+	username := NewUsername(auth.Config.Username)
+
+	// Generate CSR
```
Tovli: why is this comment necessary?

pkg/authenticator/authenticator_test.go
```diff
@@ -48,4 +52,32 @@ func TestAuthenticator(t *testing.T) {
 			So(authn.IsCertExpired(), ShouldEqual, true)
 		})
 	})
+
+	Convey("GenerateCSR", t, func() {
```
Tovli: Are those 3 different tests in one?
Why not creating 3 different tests?

pkg/authenticator/username.go
```diff
@@ -0,0 +1,35 @@
+package authenticator
+
+import "strings"
+
+// Represents the username of the host that is authenticating with Conjur.
```
Tovli: Seems better to put this in the ticket

pkg/authenticator/username_test.go
```diff
@@ -0,0 +1,24 @@
+package authenticator
+
+import (
+	"testing"
+
+	. "github.com/smartystreets/goconvey/convey"
+)
+
+func TestUsername(t *testing.T) {
```
Tovli: Is there a standard for tests names? Like UsernameSplittedSuccessfully ?
What can I know if I see TestUsername failed in the output of the tests?

pkg/authenticator/authenticator.go
orenbm: This comment may be also in the ticket but we also have it in the code for the next reader who will never look for a ticket.

Regarding the PR name - we do not have the ticket in it but we do have it in the description. I'll add it.

pkg/authenticator/authenticator.go
orenbm: It is not. I will remove it.

pkg/authenticator/authenticator_test.go
orenbm: it is one test only. 
`Convey("GenerateCSR"` states the beginning of the UT for `GenerateCSR`. Then we have the preparation, and then we have 1 `Given` clause and 2 `Then` clauses which test 2 outcomes:
1. The method finishes without an error
2. The common-name is inserted in the certificate subject as expected.

pkg/authenticator/username.go
orenbm: Again, this improves code readability.

pkg/authenticator/username_test.go
orenbm: Here is when the Convey clauses below come to action. For example, you will see a failure in:
`NewUsername -> Given a host's username -> Splits the username as expected` and will understand the context and the unexpected output.

pkg/authenticator/username_test.go
```diff
@@ -0,0 +1,24 @@
...
+	Convey("NewUsername", t, func() {
```
nessiLahav: I think we should elaborate the testing here, and to understand which behavior we are expecting, for example:
1. username = string1
2. username = string1/string2

pkg/authenticator/requests_test.go
```diff
@@ -0,0 +1,28 @@
...
+)
+
+func TestRequests(t *testing.T) {
+	Convey("LoginRequest", t, func() {
+		conjurVersion := "5"
+		authnURL := "dummyURL"
+		csrBytes := []byte("dummyCSRBytes")
+
+		Convey("Given a host's username prefix", func() {
+			usernamePrefix := "host.path.to.policy"
```
nessiLahav: Do you see any value to test also an empty Prefix ?

pkg/authenticator/username_test.go
orenbm: thanks, added more tests

pkg/authenticator/requests_test.go
orenbm: we added input validation on the username that it includes a `host/` prefix so we don't have an empty prefix scenario

PR #62, opened 2019-12-29
Bump to version 1.0.0

orenbm:
The signature of NewFromEnv does not get `clientCertPath` & `TokenPath`
as parameters anymore and has default values for them instead. This is
an incompatible API change and thus we need to bump the MAJOR version.


orenbm:
Closing this PR. we opened #63 instead.

PR #67, opened 2020-01-21
Bump version to v0.16.0

orenbm:
This version enables authenticating hosts that have their application
identity defined in annotations instead of in the id.


VERSION
```diff
@@ -1 +1 @@
-0.15.0
+0.16.0
```
sigalsax: seems the version might be out of order? See here: https://github.com/cyberark/conjur-authn-k8s-client/pull/62/files  we pumped to 1.0.0 and now we are returning to 0.16.0?

sigalsax:
approved, merge whenever you are ready

VERSION
orenbm: we didn't bump to 1.0.0 in the end.

Issue #69, opened 2020-01-29
Client publishes Docker images only when new version is tagged


izgeri:
At current, the client will publish new Docker images for every commit that hits master / every daily run of the master pipeline. New images should be pushed to DockerHub and RedHat only when the version tag changes, however.

AC:
- [x] `./bin/publish` is refactored to publish only when the master branch commit matches that of a github tag (the same way that [secretless does](https://github.com/cyberark/secretless-broker/blob/master/bin/publish))

Follow-up issue: #71

JakeQuilty:
https://github.com/cyberark/conjur-authn-k8s-client/pull/72

PR #72, opened 2020-02-04
Update publish to only push to Docker Hub on new version

JakeQuilty:
Used https://github.com/cyberark/secretless-broker/commit/56462fbfb64f89900d3ccd8a707e5a5b1273a6e1#diff-66a2901e5a92803e08f503c72520e8a8 as a reference.

Updated `publish` file to only push to Docker Hub if the tag matches `VERSION`. Used an old Secretless commit as a reference. Tested locally, but haven't been able to test to make sure it actually pushes to Docker Hub. 

Issue: https://github.com/cyberark/conjur-authn-k8s-client/issues/69

### DIFF
--- a/bin/build_utils
+++ b/bin/build_utils
@@ -1,0 +1,21 @@
+#!/bin/bash
+
+set -euo pipefail
+
+####
+# Functions to generate version numbers for this project
+####
+
+function version_tag() {
+   echo "$(< VERSION)-$(git rev-parse --short HEAD)"
+}
+
+# generate less specific versions, eg. given 1.2.3 will print 1.2 and 1
+# (note: the argument itself is not printed, append it explicitly if needed)
+function gen_versions() {
+  local version=$1
+  while [[ $version = *.* ]]; do
+    version=${version%.*}
+    echo $version
+  done
+}
--- a/bin/publish
+++ b/bin/publish
@@ -1,24 +1,54 @@
 #!/bin/bash -e
 
-DOCKERHUB_IMAGE_OLD='cyberark/conjur-kubernetes-authenticator'  # backwards-compatible image name
-DOCKERHUB_IMAGE_NEW='cyberark/conjur-authn-k8s-client'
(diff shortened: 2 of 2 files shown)


bin/publish
```diff
@@ -1,24 +1,51 @@
...
+      docker push "$REGISTRY/$image_name:$tag"
+    done
+  fi
+done
 
 docker tag conjur-authn-k8s-client:dev-redhat "$REDHAT_IMAGE"
-docker tag conjur-authn-k8s-client:dev-redhat "$REDHAT_IMAGE:$VERSION_TAG"
+docker tag conjur-authn-k8s-client:dev-redhat "$REDHAT_IMAGE:$VERSION"
+
+docker push "$REDHAT_IMAGE:$VERSION" || { echo 'Red Hat push FAILED!'; exit 0; }  # you can't push the same tag twice to redhat registry, so ignore errors
```
BradleyBoutcher: I'd like @sgnn7 's input as well, but my only nit here is the several different methods used for string interpolation. If my memory serves, we always try to use this format:
"${variable}"

bin/build_utils
```diff
@@ -0,0 +1,21 @@
...
+
+# generate less specific versions, eg. given 1.2.3 will print 1.2 and 1
+# (note: the argument itself is not printed, append it explicitly if needed)
+function gen_versions() {
+  local version=$1
+  while [[ $version = *.* ]]; do
+    version=${version%.*}
+    echo $version
+  done
+}
```
doodlesbykumbi: nit: add newline at EOF

bin/publish
sgnn7: Yeah this one should be enclosed in braces since it seems like it would break things otherwise. I think also in this case, a real `if` would be easier to read/maintain:
```bash
if ! docker push "${REDHAT_IMAGE}:${VERSION}"; then
  echo 'Red Hat push FAILED!'
  exit 0
fi
```

JakeQuilty: Just changed that

PR #74, opened 2020-02-06
Fixed tag referring to an image that doesn't exist

JakeQuilty:


izgeri:
Quick q - I saw this in my gh notifications, and I'm wondering why we always push the RH image (and don't just push when a new version is tagged, like everything else)

sgnn7:
@izgeri sadly, I don't know enough about RH images to answer that but the original file had it the same way (I also see someone's familiar name on that commit 😺) : https://github.com/cyberark/conjur-authn-k8s-client/blob/870e2f44b29474d3d6aaa1d6b922efb2529b63ad/bin/publish#L24

izgeri:
Oh dear :disappear:

I think we should update this to only push on new tags

I don't think it will functionally be different, bc RH evidently doesn't allow overwriting existing tags, but it's clearer on reading the file later and more consistent with how we handle the other images

PR #80, opened 2020-02-12
Default CONJUR_VERSION behavior refactored and tested

BradleyBoutcher:
When building a configuration from the environment, we now verify
that the input value for the Conjur version is '4' or '5'. If empty,
it will use a default value of '5'. If any other value is provided, we
throw an error.

- Performed a minor refactoring of NewFromEnv to allow functional options as arguments
- Created unit tests for conjur versions, with a framework which allows for additional unit tests for any of the features in config.go

### DIFF
--- a/pkg/authenticator/config/config.go
+++ b/pkg/authenticator/config/config.go
@@ -1,6 +1,7 @@
 package config
 
 import (
+	"fmt"
 	"io/ioutil"
 	"os"
 	"time"
@@ -24,95 +25,59 @@ type Config struct {
 	Username            *Username
 }
 
+// Default settings (this comment added to satisfy linter)
 const (
 	DefaultClientCertPath = "/etc/conjur/ssl/client.pem"
 	DefaultTokenFilePath  = "/run/conjur/access-token"
+
+	DefaultConjurVersion = "5"
 
 	// DefaultTokenRefreshTimeout is the default time the system waits to reauthenticate on error
 	DefaultTokenRefreshTimeout = 6 * time.Minute
 )
 
-// New returns a new authenticator configuration object
+var requiredEnvVariables = []string{
+	"CONJUR_AUTHN_URL",
+	"CONJUR_ACCOUNT",
+	"CONJUR_AUTHN_LOGIN",
+	"MY_POD_NAMESPACE",
+	"MY_POD_NAME",
+}
+
+// ReadFileFunc defines the interface for reading an SSL Certificate from the env
+type ReadFileFunc func(filename string) ([]byte, error)
+
+// NewFromEnv returns a config FromEnv using the standard file reader for reading certs
 func NewFromEnv() (*Config, error) {
+	return FromEnv(ioutil.ReadFile)
+}
+
+// FromEnv returns a new authenticator configuration object
+func FromEnv(readFileFunc ReadFileFunc) (*Config, error) {
 	var err error
 
-	// Check that required environment variables are set
-	for _, envvar := range []string{
-		"CONJUR_AUTHN_URL",
-		"CONJUR_ACCOUNT",
-		"CONJUR_AUTHN_LOGIN",
-		"MY_POD_NAMESPACE",
-		"MY_POD_NAME",
-	} {
-		if os.Getenv(envvar) == "" {
-			return nil, log.RecordedError(log.CAKC009E, envvar)
-		}
-	}
-
-	// Load CA cert
-	caCert, err := readSSLCert()
-	if err != nil {
-		return nil, log.RecordedError(log.CAKC021E, err.Error())
-	}
-
-	// Load configuration from the environment
-	authnURL := os.Getenv("CONJUR_AUTHN_URL")
-	account := os.Getenv("CONJUR_ACCOUNT")
-	authnLogin, err := NewUsername(os.Getenv("CONJUR_AUTHN_LOGIN"))
+	// Fill config with 'simple' values from environment
+	config, err := populateConfig()
 	if err != nil {
 		return nil, err
 	}
 
-	podNamespace := os.Getenv("MY_POD_NAMESPACE")
-	podName := os.Getenv("MY_POD_NAME")
-
-	containerMode := os.Getenv("CONTAINER_MODE")
-
-	conjurVersion := os.Getenv("CONJUR_VERSION")
-	if len(conjurVersion) == 0 {
-		conjurVersion = "5"
+	// Load CA cert from Environment
+	config.SSLCertificate, err = readSSLCert(readFileFunc)
+	if err != nil {
+		return nil, log.RecordedError(log.CAKC021E, err.Error())
 	}
 
-	// Parse token refresh rate if one is provided from env
-	tokenRefreshTimeout := DefaultTokenRefreshTimeout
-	tokenRefreshTimeoutString := os.Getenv("CONJUR_TOKEN_TIMEOUT")
-	if len(tokenRefreshTimeoutString) > 0 {
-		parsedTokenRefreshTimeout, err := time.ParseDuration(tokenRefreshTimeoutString)
-		if err != nil {
-			return nil, log.RecordedError(log.CAKC010E, err.Error())
-		}
-
-		tokenRefreshTimeout = parsedTokenRefreshTimeout
+	// Load Username from Environment
+	config.Username, err = NewUsername(os.Getenv("CONJUR_AUTHN_LOGIN"))
+	if err != nil {
+		return nil, err
 	}
 
-	tokenFilePath := DefaultTokenFilePath
-	// If CONJUR_TOKEN_FILE_PATH is defined in the env we take its value
-	if envVal := os.Getenv("CONJUR_AUTHN_TOKEN_FILE"); envVal != "" {
-		tokenFilePath = envVal
-	}
-
-	clientCertPath := DefaultClientCertPath
-	// If CONJUR_CLIENT_CERT_PATH is defined in the env we take its value
-	if envVal := os.Getenv("CONJUR_CLIENT_CERT_PATH"); envVal != "" {
-		clientCertPath = envVal
-	}
-
-	return &Config{
-		Account:             account,
-		ClientCertPath:      clientCertPath,
-		ContainerMode:       containerMode,
-		ConjurVersion:       conjurVersion,
-		PodName:             podName,
-		PodNamespace:        podNamespace,
-		SSLCertificate:      caCert,
-		TokenFilePath:       tokenFilePath,
-		TokenRefreshTimeout: tokenRefreshTimeout,
-		URL:                 authnURL,
-		Username:            authnLogin,
-	}, nil
+	return config, nil
 }
 
-func readSSLCert() ([]byte, error) {
+func readSSLCert(readFile ReadFileFunc) ([]byte, error) {
 	SSLCert := os.Getenv("CONJUR_SSL_CERTIFICATE")
 	SSLCertPath := os.Getenv("CONJUR_CERT_FILE")
 	if SSLCert == "" && SSLCertPath == "" {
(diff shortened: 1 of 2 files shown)


pkg/authenticator/config/config.go
```diff
@@ -28,12 +29,37 @@ const (
...
+}
+
+// DefaultNewFromEnv returns NewFromEnv with the default functional options
+func DefaultNewFromEnv() (*Config, error) {
+	return NewFromEnv(DefaultConfigOptions)
+}
+
 // New returns a new authenticator configuration object
-func NewFromEnv() (*Config, error) {
+func NewFromEnv(opts Options) (*Config, error) {
```
sgnn7: This is a breaking change in almost all packages that use our authenticator. I would highly avoid changing this method's signature. As far as I can tell, this was changed to accommodate the tests but this can be tested without this change I think.

pkg/authenticator/config/config.go
```diff
@@ -28,12 +29,37 @@ const (
...
+}
+
+// DefaultConfigOptions implement the methods defined within config.go
+var DefaultConfigOptions = Options{
+	readSSLCert,
+	NewUsername,
+}
+
+// DefaultNewFromEnv returns NewFromEnv with the default functional options
+func DefaultNewFromEnv() (*Config, error) {
```
sgnn7: Normally, the prefix for Golang constructors is `New` so if a change is needed, it should contain methods similar to `NewFromEnv` and `NewFromEnvWithOptions`.

pkg/authenticator/config/config.go
```diff
@@ -28,12 +29,37 @@ const (
...
+type Options struct {
+	ReadSSLCert ReadSSLCertFunc
+	NewUserName NewUserNameFunc
+}
+
+// DefaultConfigOptions implement the methods defined within config.go
+var DefaultConfigOptions = Options{
+	readSSLCert,
+	NewUsername,
+}
```
sgnn7: I could be wrong but I would double-check that you really need any of these types to blackbox-test the authenticator - I think they might not be needed.

pkg/authenticator/config/config.go
```diff
@@ -68,9 +94,12 @@ func NewFromEnv() (*Config, error) {
...
-	conjurVersion := os.Getenv("CONJUR_VERSION")
-	if len(conjurVersion) == 0 {
-		conjurVersion = "5"
+	conjurVersion := DefaultConjurVersion
+	// Only '4' and '5' are supported, with empty values using the default of '5'
+	if version := os.Getenv("CONJUR_VERSION"); version == "4" {
+		conjurVersion = version
+	} else if version != "5" && version != "" {
+		return nil, log.RecordedError(log.CAKC021E, fmt.Errorf("invalid conjur version"))
 	}
```
sgnn7: This feels like it should be a `select` rather than an `if`

pkg/authenticator/config/config_test.go
```diff
@@ -0,0 +1,80 @@
...
+	"github.com/cyberark/conjur-authn-k8s-client/pkg/log"
+	"os"
+	"testing"
+
+	. "github.com/smartystreets/goconvey/convey"
+)
+
+var mockSuccessfulOptions = Options{
+	successfulMockReadSSLCertFunc,
+	 successfulMockNewUserNameFunc,
```
sgnn7: nit: Indentation is inconsistent

pkg/authenticator/config/config_test.go
```diff
@@ -0,0 +1,80 @@
...
+
+var environmentValues = map[string]string{
+	"CONJUR_SSL_CERTIFICATE": "testSSLCert",
+	"CONJUR_CERT_FILE": "testSSLCertFile.txt",
+	"CONJUR_AUTHN_URL": "filepath",
+	"CONJUR_ACCOUNT": "testAccount",
+	"CONJUR_AUTHN_LOGIN": "host",
+	"MY_POD_NAMESPACE": "testNameSpace",
+	"MY_POD_NAME": "testPodName",
+	"CONJUR_VERSION": "",
```
sgnn7: nit: Would be good to sort these since it's hard to find a variable in an unordered list

pkg/authenticator/config/config_test.go
```diff
@@ -0,0 +1,80 @@
...
+}
+
+func TestAuthenticator(t *testing.T) {
+	// Set default environment variables
+	for key, value := range environmentValues {
+		err := os.Setenv(key, value)
+		if err != nil {
+			panic(err)
+		}
+	}
```
sgnn7: This should have a `defer`ed unsetting of those variables to not pollute other tests

pkg/authenticator/config/config.go
```diff
@@ -68,9 +94,12 @@ func NewFromEnv() (*Config, error) {
 
 	containerMode := os.Getenv("CONTAINER_MODE")
 
-	conjurVersion := os.Getenv("CONJUR_VERSION")
-	if len(conjurVersion) == 0 {
-		conjurVersion = "5"
+	conjurVersion := DefaultConjurVersion
+	// Only '4' and '5' are supported, with empty values using the default of '5'
+	if version := os.Getenv("CONJUR_VERSION"); version == "4" {
+		conjurVersion = version
```
izgeri: I understand that this gives the intended result (the tests make that clear, if nothing else)

but the logic seems odd to me, because it is not clear that when CONJUR_VERSION is set to 5 we also set version = CONJUR_VERSION

I know it is simpler the way you have it written, and achieves the same end result - but I feel it could be clearer what the intended behavior is, even if it ends up being a bit more verbose.

but not a dealbreaker, in any case

pkg/authenticator/config/config.go
BradleyBoutcher: The issue lies mostly with readSSLCert requiring an actual file to read, and NewUsername is also an external method which requires a specific value, which is difficult to mock out well. If possible I felt this functional option implementation would make it easier to test failure paths in NewFromEnv and to add tests for the other code paths while maintaining readability. Of course, I very much don't want to break existing functionality, but I'll need help figuring out the best way to do this.

BradleyBoutcher: For now I just swapped the signatures in a quick fix, but I definitely want to hear more about better ways we could do this

pkg/authenticator/config/config.go
diverdane: I agree, the logic would be clearer and more extensible to future versions with a switch, something like this (leave out the comments if it's too verbose):
```
	conjurVersion := os.Getenv("CONJUR_VERSION")
	switch conjurVersion {
	case "4": // Valid
	case "5": // Valid
	case "":  // Use default
		conjurVersion = DefaultConjurVersion
	default: // Unknown version
		return nil, log.RecordedError(log.CAKC021E, fmt.Errorf("invalid conjur version"))
	}
```

pkg/authenticator/config/config_test.go
```diff
@@ -0,0 +1,80 @@
...
+
+	Convey("NewFromEnv", t, func() {
```
diverdane: This is a matter of personal preference, but table driven test could be used here to consolidate the corner cases that are considered, e.g.:
```
func TestAuthenticator(t *testing.T) {
	// Set default environment variables
	for key, value := range environmentValues {
		err := os.Setenv(key, value)
		if err != nil {
			panic(err)
		}
	}

	TestCases := []struct {
		description string
		envVersion  string
		expVersion  string
		expErrStr   string
	}{
		{
			description: "Succeeds if version is 4",
			envVersion:  "4",
			expVersion:  "4",
			expErrStr:   "",
		},
		{
			description: "Succeeds if version is 5",
			envVersion:  "5",
			expVersion:  "5",
			expErrStr:   "",
		},
		{
			description: "Sets the default version for an empty value",
			envVersion:  "",
			expVersion:  DefaultConjurVersion,
			expErrStr:   "",
		},
		{
			description: "Returns error if version is invalid",
			envVersion:  "3",
			expVersion:  "",
			expErrStr:   fmt.Sprintf(log.CAKC021E, "invalid conjur version"),
		},
	}

	Convey("NewFromEnv", t, func() {
		for _, tc := range TestCases {
			Convey(tc.description, func() {
				_ = os.Setenv("CONJUR_VERSION", tc.envVersion)

				config, err := NewNewFromEnv(mockSuccessfulOptions)

				if tc.expErrStr == "" {
					So(err, ShouldBeNil)
					So(config.ConjurVersion, ShouldEqual, tc.expVersion)
				} else {
					So(err, ShouldNotBeNil)
					So(err.Error(), ShouldEqual, tc.expErrStr)
				}
			})
		}
	})
}
```

pkg/authenticator/config/config.go
```diff
@@ -68,9 +99,17 @@ func NewFromEnv() (*Config, error) {
...
 	containerMode := os.Getenv("CONTAINER_MODE")
 
-	conjurVersion := os.Getenv("CONJUR_VERSION")
-	if len(conjurVersion) == 0 {
-		conjurVersion = "5"
+	// Only versions '4' & '5' are allowed, with '5' being used as the default
+	conjurVersion := DefaultConjurVersion
+	switch os.Getenv("CONJUR_VERSION") {
+	case "4":
+		conjurVersion = os.Getenv("CONJUR_VERSION")
```
diverdane: A slight variation, maybe... Instead of calling os.Getenv() twice, you already know what it's going to be:
```
	case "4":
		conjurVersion = "4"
```

pkg/authenticator/config/config.go
```diff
@@ -28,37 +29,67 @@ const (
...
+	NewUserName NewUserNameFunc
+}
+
+// DefaultConfigOptions implement the methods defined within config.go
+var DefaultConfigOptions = Options{
+	readSSLCert,
+	NewUsername,
+}
+
+// NewFronEnv returns a config FromEnv with the default functional options
```
diverdane: Typo, "NewFronEnv"

pkg/authenticator/config/config.go
```diff
@@ -28,37 +29,67 @@ const (
...
+	caCert, err := opts.ReadSSLCert()
 	if err != nil {
 		return nil, log.RecordedError(log.CAKC021E, err.Error())
 	}
 
 	// Load configuration from the environment
 	authnURL := os.Getenv("CONJUR_AUTHN_URL")
 	account := os.Getenv("CONJUR_ACCOUNT")
-	authnLogin, err := NewUsername(os.Getenv("CONJUR_AUTHN_LOGIN"))
+	authnLogin, err := opts.NewUserName(os.Getenv("CONJUR_AUTHN_LOGIN"))
```
diverdane: I believe that the `NewUserName` call here doesn't need to be mocked for UT. To convince myself, I put back the original call here on line 92, and the UT ran fine.

pkg/authenticator/config/config.go
```diff
@@ -28,88 +29,109 @@ const (
...
+	"CONJUR_AUTHN_LOGIN",
+	"MY_POD_NAMESPACE",
+	"MY_POD_NAME",
+}
+
+// NewFromEnvFunc defines the method for creating a new config from the env
+type NewFromEnvFunc func(Options) (*Config, error)
+
+// ReadSSLCertFunc defines the method for reading an SSL Certificate from the env
+type ReadSSLCertFunc func() ([]byte, error)
```
diverdane: Instead of mocking the `readSSLCert` function, you could get better coverage by mocking the lower level file reader inside the `readSSLCert` function itself, i.e. this line:
```
return ioutil.ReadFile(SSLCertPath)
```
The way to do this would be to define a `fileReader` interface, and add a `fileReader` interface as an argument to the `FromEnv()` function signature (no need to use functional options since there's only one option, and it's a simple interface). Within `FromEnv()`, you would pass on that interface to the call to `readSSLCert()`, and inside `readSSLCert()`, you would do the reading of the file using the interface passed in.
Then in your UT, you would pass in a buffer or some other IO reader as a mock.

pkg/authenticator/config/config.go
```diff
@@ -28,88 +29,109 @@ const (
...
+	config.SSLCertificate, err = opts.ReadSSLCert()
 	if err != nil {
 		return nil, log.RecordedError(log.CAKC021E, err.Error())
 	}
 
 	// Load configuration from the environment
-	authnURL := os.Getenv("CONJUR_AUTHN_URL")
-	account := os.Getenv("CONJUR_ACCOUNT")
-	authnLogin, err := NewUsername(os.Getenv("CONJUR_AUTHN_LOGIN"))
+	config.Username, err = opts.NewUserName(os.Getenv("CONJUR_AUTHN_LOGIN"))
```
diverdane: Apologies if this comment shows up twice, but I didn't see it showing up in the file diffs:
I believe that the NewUserName call here doesn't need to be mocked for UT. To convince myself, I put back the original call here on line 92, and the UT ran fine.

pkg/authenticator/config/config_test.go
```diff
@@ -0,0 +1,95 @@
...
+					So(err.Error(), ShouldEqual, tc.expErrStr)
+				}
+			})
+		}
+	})
+}
+
+func successfulMockReadFile(filename string) ([]byte, error) {
+	return []byte{}, nil
+}
```
diverdane: Github complains: Missing newline on line 95.

pkg/authenticator/config/config.go
```diff
@@ -28,91 +29,94 @@ const (
...
+
+// NewFromEnv returns a config FromEnv with the default functional options
 func NewFromEnv() (*Config, error) {
-	var err error
+	return FromEnv(ioutil.ReadFile)
+}
+
+// FromEnv returns a new authenticator configuration object
+func FromEnv(readFileFunc ReadFileFunc) (*Config, error) {
+	var err error = nil
```
diverdane: Can drop the nil, since that's the zero value for an error type, i.e. just use `var err error`.

pkg/authenticator/config/config.go
```diff
@@ -28,91 +29,94 @@ const (
 	DefaultClientCertPath = "/etc/conjur/ssl/client.pem"
```
diverdane: Some linters are going to complain about `DefaultClientCertPath` not having a comment for an exported variable. Same with the other Default* consts below. You could satisfy linter for all of these with a comment before the `const (`, e.g. :
```
// Default settings (this comment added to satisfy linter)
const (
```

pkg/authenticator/config/config.go
```diff
@@ -28,91 +29,94 @@ const (
...
+	"CONJUR_ACCOUNT",
+	"CONJUR_AUTHN_LOGIN",
+	"MY_POD_NAMESPACE",
+	"MY_POD_NAME",
+}
+
+// ReadFileFunc defines the interface for reading an SSL Certificate from the env
+type ReadFileFunc func(filename string) ([]byte, error)
+
+// NewFromEnv returns a config FromEnv with the default functional options
```
diverdane: "with the default functional options" should be "using the standard file reader" or something like that.

pkg/authenticator/config/config.go
```diff
@@ -25,102 +26,124 @@ type Config struct {
 }
 
 const (
+	// DefaultClientCertPath (this comment added to satisfy linter)
```
diverdane: You could have used one comment at the head of this block (before the `const (` line) to handle all of the exported consts, but this works.

BradleyBoutcher:
@sgnn7 
Done

Issue #85, opened 2020-02-17
On successful run k8s-secrets-provider build will trigger


eladkug:
Set trigger for build automation so successful run will trigger k8s-secrets-provider build
DOD:

 - [x] Run successful build and verify it trigger new build
- [x] Run unsuccessful build and verify there is no trigger

eladkug:
Decided not to implement since its not relevant.

PR #86, opened 2020-02-17
 trigger cyberark--secrets-provider-for-k8s build on success

eladkug:
What does this PR do?
If build success it will trigger cyberark--secrets-provider-for-k8s and will not wait for the result.

What ticket does this PR close?
#85 

Verified on Master as well.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -34,8 +34,15 @@ pipeline {
   }
 
   post {
-    always {
-      cleanupAndNotify(currentBuild.currentResult)
-    }
+      success {
+          script {
+               if (env.BRANCH_NAME == 'master') {
+                      build (job:'../cyberark--secrets-provider-for-k8s/master', wait: false)
+               }
+          }
+      }
+      always {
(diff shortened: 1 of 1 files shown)


eladkug:
> Code looks good, but why do you want conjur-authn-k8s-client to trigger secrets-provider-for-k8s?

I want to verify any change on authn-k8s will not influence on my tests as well  (for example conjur-authn_failure verification)

hughsaunders:
@eladkug In that case, why not pin the version of authn-k8s-client, then changes won't cause your tests to fail. You can have Jenkins regularly (eg daily) check for new versions and create a pull request against your own repo that will update to the latest version of authn-k8s-client, and if the tests on that PR succeed, merge it, if not communicate between the two teams to fix it. 

This method gives you a notification that a new verison is out (via the PR), and an opportunity to run tests against the new version without breaking master.

eladkug:
Thanks for your input i will verify this again and if its irrelevant i will close the PR

eladkug:
Decided to close this PR. Thanks for your inputs

PR #94, opened 2020-03-30
Adding trivy scanning

andytinkham:
Adds trivy scanning with no build failure to pipeline
Signed-off-by: Andy Tinkham <andy.tinkham@cyberark.com>

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -26,8 +26,16 @@ pipeline {
     stage('Run Tests') {
       steps {
         sh './bin/test'
-        
+
         junit 'test/junit.xml'
+      }
+    }
+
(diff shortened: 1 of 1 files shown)


andytinkham:
I can file an issue - @ismarc indicated there's very little that trivy would catch even if we could scan it because we're just copying binaries into the empty container. Matt, do I have that right?

izgeri:
hash tag super secure :success_kid:

ismarc:
> I can file an issue - @ismarc indicated there's very little that trivy would catch even if we could scan it because we're just copying binaries into the empty container. Matt, do I have that right?

Yup, that's correct.  Because it's `FROM scratch`, there's no installed packages or anything to check versions of, just copied in binaries, but it has a substantially smaller surface area.

Issue #91, opened 2020-03-05
FIPS compliance for authn-client-k8s


hilagross:
# FIPS compliance for authn-client-k8s

### Overview
The National Institute of Standards and Technology (NIST) issued the [FIPS 140-2](https://csrc.nist.gov/publications/detail/fips/140/2/final) Publication Series to coordinate the requirements and standards for cryptography modules that include both hardware and software components. 

The requirements are to support FIPS in all cryptography 3rd parties, preferably as one source code if possible, otherwise by configuration. 

As making Conjur product FIPS compliant, our costumer will gain a cryptographic standard to protect their unclassified, but sensitive data.

As part of the effort of making Conjur FIPS compliant, we tackled an issue regarding Golang. 
The issue is that Golang, by nature, is not FIPS compliant and they claim that they are not going to be (https://github.com/golang/go/issues/11658#issuecomment-120441723). 

There are three alternatives:
1. Use cgo to call out to an existing certified library
   - This alternative disqualified due to its complexity and consuming time and effort as we will have to replace any usage of crypto.

2. Use BoringSSL based crypto
   - Use [Google’s fork of Golang](https://github.com/golang/go/blob/dev.boringcrypto.go1.12/misc/boring/README.md) that uses dev.boringcrypto as its crypto engine. dev.boringcrypto wraps BoringSSL which is FIPS compliant.
   - disadvantage of this approach is that Google does not guarantee any “support” or responsibility on this version being FIPS compliant.
3. Use [RHEL go toolchain](https://developers.redhat.com/blog/2019/06/24/go-and-fips-140-2-on-red-hat-enterprise-linux/).
   - RHEL takes the ownership on the encryption, and bridges between the Golang encryption and OpenSSL.
   - What needs to be done is build the Golang project on subscribed RHEL machine/UBI container with the relevant go-toolset installed on it. than, we can take the compiled binary and copy it into the containers as usual.
   - A big advantage of this approach is that we are aligning with Red Hat which is one of the big players in the field of certifications, enterprise, and federal market. 
   - The pipeline should use and subscribe RHEL / UBI on-the-fly.

### Requirement

Compare all options (cgo, google toolset, RHEL) and decide on the best approach. should include also in seed-fetcher container.
Consider all 3 alternatives:
1. Use cgo to call out to an existing certified library
2. Use BoringSSL based crypto.
3. Use RHEL go toolchain.
Create a test plan and design the solution before implementing it.

**our goal:**
[...]

PR #97, opened 2020-04-01
Build and publish a FIPS compliant image of conjur-authn-k8s-client

orenbm:
Connected to #91 

We want to let customers who require that the binary is built using
FIPS-compliant libraries an image of the conjur-authn-k8s-client.

This commit changes the base image from `golang` to `goboring/golang` and makes the required changes for it to work.

The image names stay the same as before.

It's important to state that the performance of the FIPS-compliant version is the same as the original one, as can be seen [here](https://github.com/cyberark/conjur-authn-k8s-client/pull/112).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM golang:1.12 as authenticator-client-builder
+FROM goboring/golang:1.12.17b4 as authenticator-client-builder
 MAINTAINER CyberArk Software Ltd.
 
 ENV GOOS=linux \
@@ -23,6 +23,10 @@ RUN go get -u github.com/jstemmer/go-junit-report && \
 RUN go build -a -installsuffix cgo \
     -ldflags="-X github.com/cyberark/conjur-authn-k8s-client/pkg/authenticator.Tag=$TAG" \
     -o authenticator ./cmd/authenticator
+
+# Verify the binary is using BoringCrypto.
+# Outputting to /dev/null so the output doesn't include all the files
+RUN sh -c "go tool nm authenticator | grep '_Cfunc__goboringcrypto_' 1> /dev/null"
 
 # =================== BUSYBOX LAYER ===================
 # this layer is used to get binaries into the main container
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Added
 - Authenticator client prints its version upon startup (#93)
 
+### Changed
+- The project now uses `goboring/golang` as its base image to be FIPS compliant
+
 ## [0.16.1] - 2020-02-18
 ### Fixed
 - Only publish to DockerHub / RH registry when there is a new version (#72, #74,


orenbm:
@izgeri @hilagross @shaharglazner I may be jumping the gun here but this seems to be a good solution to what we need. 

It is very simple, creates an image that is FIPS compliant and passes the tests in `kubernetes-conjur-demo` (which I ran locally). We will need of course to add another "test" in that repo once this PR is merged.

bin/build
```diff
@@ -29,4 +29,9 @@ docker build --build-arg VERSION="$VERSION" \
              -t "${TAG}-redhat" \
              .
 
+docker build --build-arg VERSION="$VERSION" \
+             -f Dockerfile.fips \
+             -t "${TAG}-fips" \
+             .
```
izgeri, line 35: this isn't going to result in the outcome you're looking for

we should probably revise the pipeline here so that we have a single Dockerfile for the current authenticator image and the RH image (see [the secretless Dockerfile here](https://github.com/cyberark/secretless-broker/blob/master/Dockerfile))

then the FIPS image Dockerfile can be a version of that where the `authenticator-builder` base image is `goboring/golang:1.12.17b4` (and the 3rd RH image reference is dropped, unless we also want a FIPS RH image)

what it's doing right now is:
- building the binary builder image based on Dockerfile
- running the binary builder image to build the binary and output it locally
- building the dockerhub image from Dockerfile.scratch (which copies the binary in)
- building the redhat image from Dockerfile.redhat (which copies the binary in)
- building the FIPS image (which has no binary once it's built, since it is not copied - it would have to be more like Dockerfile.scratch for that to work)

I'd love to see this become two dockerfiles, Dockerfile and Dockerfile.fips, that include multi-stage builds - eg each defines a builder image based on the proper golang version, and then builds the final dockerhub and RH images

lmk if you have questions about what I mean by this

hilagross:
I'm not sure I understand, for now our goal is to have only one source of truth, only FIPS compliant mode. 
Later, if there will be a performance degradation we will add a flag that can switch between 2 modes (FIPS and non-FIPS), and that we will achieve in another FIPS task.

shaharglazner:
@izgeri @orenbm 
I agree with Geri (if I understood correctly) that the `authenticator-builder` needs to build the fips compliant artifact, and then in the `Dockerfile.fips` it should be copied into the base image.

As I see it, we will have two builders (let's say `golang-builder` and `golang-fips-builder`), where every Golang project that we manage will use one of them (or both) according to its needs.

orenbm:
> for now our goal is to have only one source of truth, only FIPS compliant mode.

@hilagross I agree with @izgeri that we still need to provide a version of the authn-client which will have the original `golang` libraries. Is there a problem to provide a version of the authn-client that is FIPS compliant which customers can use?, alongside one that doesn't?

orenbm:
@hilagross there is also a good question raised by @izgeri - do we want also the Red Hat image to be FIPS compliant?

orenbm:
@izgeri I fixed the code so it does what I intend to do in this PR. please re-review :)

orenbm:
Open questions:
  1. Do we want to have 2 versions in the main registry - one that is FIPS compliant and one that is not (whether the base image is `golang` or `goboring/golang`?
  2. Do we want to have 2 versions in the *Red Hat* registry - one that is FIPS compliant and one that is not (whether the base image is `golang` or `goboring/golang`?

bin/publish
```diff
@@ -1,63 +1,87 @@
 #!/bin/bash -e
```
izgeri, line 1: I don't find the reorganization of this file to be more clear - in fact, it's more complicated imo, and jumps around quite a bit to follow the logic.

Some notes:
- For this project nothing is going to publish unless the condition `[ "$git_description" = "v${VERSION}" ]` is met, so there should be no call to a `publish_image` method outside of that condition check
- It's confusing to have `publish_image` and `publish_images` - those names are so similar, and on the surface I can't tell what is different between them except that perhaps one does the operation for more than one image
- If you really want to wrap it all up to be as simple as possible, why not do something like:
   ```
   # check that the tag matches the version
   # looping over the target images and tags (ie a double loop), publish to dockerhub
   # publish to redhat
   ```
   You can even set it up so that `publish_dockerhub_image` re-tags and pushes to dockerhub (that's really all it should need - you can pass in the base image, target, and tag) and `publish_redhat_image` does the whole login / push / error catching for a single image.
- Note: we can't push the RH FIPS image without creating an authn-k8s FIPS project in the RH container registry - so that should maybe get pushed to a follow-up ticket. We can leave the build steps in this PR, but I don't think we can publish that image yet in this PR
- I have a preference to bash methods being prefaced by `function`, but this is more of a nit - from the [style guide](https://google.github.io/styleguide/shellguide.html#function-names):
   > The function keyword is extraneous when “()” is present after the function name, but enhances quick identification of functions.
- There should be a space between the close parens of a method statement and the opening curly brace of the method definition, eg `main() {`

Dockerfile
```diff
@@ -1,9 +1,9 @@
-FROM golang:1.12 as authenticator-client-builder
+FROM goboring/golang:1.12.17b4 as authenticator-client-builder
 MAINTAINER CyberArk Software Ltd.
 
 ENV GOOS=linux
 ENV GOARCH=amd64
-ENV CGO_ENABLED=0
+ENV CGO_ENABLED=1
```
izgeri, line 6: @orenbm I'm asking the @cyberark/community-and-integrations-team to take a look at this change and the copying of libraries below to make sure we understand the implications here and that we agree the change is needed

Can you please make sure the PR description / github commit messages are clear about the motivation for the changes? It is very important for changes like this that we document exactly why we're making these choices for the benefit of community members and also for our future selves. Thanks!

Dockerfile
```diff
@@ -22,6 +22,17 @@ RUN go build -a -installsuffix cgo \
     -ldflags="-X github.com/cyberark/conjur-authn-k8s-client/pkg/authenticator.Tag=$TAG" \
     -o authenticator ./cmd/authenticator
 
+# verify the binary is using BoringCrypto.
+# Outputting to /dev/null so the output doesn't include all the files
+RUN sh -c "go tool nm authenticator | grep '_Cfunc__goboringcrypto_' 1> /dev/null"
```
sgnn7, line 27: I don't know for sure but it would be good to check if `sh` fails when grep fails too (in `bash` you need both `set -e` and `set -o pipefail` for this line to work as expected).

Dockerfile
```diff
@@ -22,6 +22,17 @@ RUN go build -a -installsuffix cgo \
     -ldflags="-X github.com/cyberark/conjur-authn-k8s-client/pkg/authenticator.Tag=$TAG" \
     -o authenticator ./cmd/authenticator
 
+# verify the binary is using BoringCrypto.
```
sgnn7, line 25: Capitalize your comment's first word
```suggestion
# Verify the binary is using BoringCrypto.
```

Dockerfile
```diff
@@ -22,6 +22,17 @@ RUN go build -a -installsuffix cgo \
...
+# Outputting to /dev/null so the output doesn't include all the files
+RUN sh -c "go tool nm authenticator | grep '_Cfunc__goboringcrypto_' 1> /dev/null"
+
+# collect dependent libraries so they're later copied to the final image
+# this is needed as we use CGO_ENABLED=1
+RUN mkdir -p lib && mkdir -p lib64 && \
```
sgnn7, line 31: All lines should be broken when they hit `&&` for easier readability

Dockerfile
```diff
@@ -22,6 +22,17 @@ RUN go build -a -installsuffix cgo \
...
+    ldd authenticator | tr -s '[:blank:]' '\n' | grep '^/' | \
+      xargs -I % sh -c 'mkdir -p $(dirname ./%); cp % ./%;' && \
```
sgnn7, line 33: This seems like it could use some simplification and error checking

Dockerfile
orenbm: yes, i verified it. the output will be:
```
Step 12/29 : RUN sh -c "go tool nm authenticator | grep '_Cfunc__goboringcrypto_' 1> /dev/null"
 ---> Running in 9711b93a98f9
The command '/bin/sh -c sh -c "go tool nm authenticator | grep '_Cfunc__goboringcrypto_' 1> /dev/null"' returned a non-zero code: 1
```

orenbm: and the build will fail of course.

orenbm:
@sgnn7 can you re-review please?

I updated the PR description and answered to your comments.

Thanks :)

PR #101, opened 2020-04-02
Remove redundant TAG variable from publish script

orenbm:
In a previous [commit](https://github.com/cyberark/conjur-authn-k8s-client/commit/19ceb52df6987f4f98049b0238640b4f69e63610) we stopped publishing our image to the internal registry. However, the TAG variable was left behind for no reason.

It was used in the script as an intermediate name for the image but it is not needed and makes the script less readable.

### DIFF
--- a/bin/publish
+++ b/bin/publish
@@ -12,8 +12,6 @@ readonly IMAGES=(
   "conjur-authn-k8s-client"
 )
 
-readonly TAG="latest"
-
 readonly TAGS=(
   "$VERSION"
   "latest"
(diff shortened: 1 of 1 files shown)


bin/publish
```diff
@@ -36,11 +32,11 @@ for image_name in "${IMAGES[@]}"; do
 
     for tag in "${TAGS[@]}" $(gen_versions $VERSION); do
       echo "Tagging and pushing $REGISTRY/$image_name:$tag"
-      docker tag "$image_name:$TAG" "$REGISTRY/$image_name:$tag"
+      docker tag conjur-authn-k8s-client:dev "$REGISTRY/$image_name:$tag"
```
sgnn7, line 35: Can you extract this `conjur-authn-k8s-client:dev` to a variable though? it would be good that we make it obvious that both images are sourced from the same source docker image which isn't too apparent here unless you go through it extremely carefully.

orenbm:
thanks @sgnn7 :)

I also added another commit with a change i noticed while implementing your request.

PR #102, opened 2020-04-04
Remove doubled push of `latest` tag

orenbm:
There was an incorrect logic where we pushed the `latest` tag twice -
once when the `for` loop was with `tag == $VERSION` and once when the
`for` loop was with `tag == latest`.

Also - renaming `IMAGES` to `DESTINATION_IMAGES` for clarity

### DIFF
--- a/bin/publish
+++ b/bin/publish
@@ -7,7 +7,9 @@ REDHAT_IMAGE='scan.connect.redhat.com/ospid-1c46a2de-1d88-40e6-a433-7114ad0099cb
 readonly REGISTRY="cyberark"
 readonly VERSION="$(short_version_tag)"
 
-readonly IMAGES=(
+# we changed the name to `conjur-authn-k8s-client`. Leaving `conjur-kubernetes-authenticator`
+# for backwards-compatibility.
+readonly DESTINATION_IMAGES=(
   "conjur-kubernetes-authenticator"
   "conjur-authn-k8s-client"
 )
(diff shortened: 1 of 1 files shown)


bin/publish
```diff
@@ -29,17 +29,12 @@ if [ "$git_description" = "v${VERSION}" ]; then
 
   echo "Pushing to Dockerhub..."
   source_image=conjur-authn-k8s-client:dev
-  # tagging and pushing to dockerhub
```
sigalsax, line 32: let's bring back this comment

bin/publish
```diff
@@ -7,7 +7,7 @@ REDHAT_IMAGE='scan.connect.redhat.com/ospid-1c46a2de-1d88-40e6-a433-7114ad0099cb
 readonly REGISTRY="cyberark"
 readonly VERSION="$(short_version_tag)"
 
-readonly IMAGES=(
+readonly DESTINATION_IMAGES=(
   "conjur-kubernetes-authenticator"
```
sigalsax, line 11: are these two images the same? just named differently? If so why do we have this duplicate?

bin/publish
orenbm: why do we need the comment if it's mentioned in the `echo` above?

bin/publish
orenbm: we changed the name from `conjur-kubernetes-authenticator` to `conjur-authn-k8s-client` but we still push the first one for backwards-compatibility. I think it is deprecated so it won't stay with us for much longer.

sigalsax: may be worth commenting that are clarity?

bin/publish
sigalsax: b/c at a glance, `$REGISTRY` could be anything (could be dockerhub or could be our Openshift registry for example). Right now it is `cyberark/conjur-authn-k8s-client:latest` but doesn't tell me it is to dockerhub so IMO it is helpful to add the comment above describing it is pushing the image to dockerhub.

orenbm: but 2 lines above it we `echo` "Pushing to Dockerhub...".

bin/publish
orenbm: Done

bin/publish
sigalsax: oh missed that. we good here then

Issue #93, opened 2020-03-27
Authenticator container writes version to log on startup


izgeri:
## Is your feature request related to a problem? Please describe.

Right now to determine authenticator version we rely on the SHA256 hashes available via `docker inspect`, but this is making it difficult to determine the exact image that was used - especially if the image wasn't taken from dockerhub directly.

## Describe the solution you would like
In cyberark/secretless-broker#944 we added our current implementation of printing the Secretless version on startup - adding a similar startup line to the authenticator container would also be nice.

Note: it should not be part of the public packages of the authenticator - adding it there would mean that every tool that is based on those packages would also print the authenticator version, which would be confusing and redundant.

AC:
- [ ] The authenticator container prints the version on startup. For tag builds, the version should print as the semver tag (eg `v1.2.3`). For dev builds, the version should print as the semver tag + the commit hash (eg `v1.2.3-{hash}`)

InbalZilberman:
@izgeri Can you please explain the note about the public packages?

izgeri:
@InbalZilberman the project has a main.go file that invokes the internal / public golang packages to start up and run the authenticator.

Secretless and the k8s secrets provider both import the public golang packages of this project to run the authenticator as part of their workflows.

Looking at the [main.go](https://github.com/cyberark/conjur-authn-k8s-client/blob/master/cmd/authenticator/main.go) for this project now, it probably makes sense to log a startup message and print the version as one of the first steps in main.go. This has the added benefit of avoiding the problem I am concerned with, which is that Secretless will start and log "Starting Secretless vx.y.z" and then log "Starting authenticator va.b.c", which I think would be a little confusing (and as I noted, redundant, since Secretless refers to the authenticator library version it's using in its NOTICES and go.mod).

Hope this makes sense, but please feel free to ask follow-up questions if any of it is not clear.

izgeri:
@orenbm this is not available in a published release yet - right? I don't think we should close it until it's available to end users, since they can't actually benefit from this change yet

orenbm:
Released a version now, thanks!

PR #107, opened 2020-04-07
Bump to version 0.17.0

orenbm:
### Added
- Added ability for authn client to print its version upon startup (#93)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -77,5 +77,13 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Fixed
 - Fix an issue where sidecar fails when not run as root user.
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -77,5 +78,14 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Fixed
 - Fix an issue where sidecar fails when not run as root user.
 
-[Unreleased]: https://github.com/cyberark/conjur-authn-k8s-client/compare/v0.13.0...HEAD
```
sigalsax, line 80: why are we bumping so many versions? 

from v0.13.0 -17?

orenbm: we are bumping only one version but i saw that the comparison section wasn't maintained. Please review this PR via the commits, it will make more sense.

PR #108, opened 2020-04-12
Improve changelog entries

orenbm:
This PR improves 2 changelog entries that now are:
  - More accurate
  - State customer facing changes instead of implementation details

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [0.17.0] - 2020-04-07
 ### Added
-- Added ability for authn client to print its version upon startup (#93)
+- Authenticator client prints its version upon startup (#93)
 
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -8,7 +8,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [0.17.0] - 2020-04-07
 ### Added
-- Added ability for authn client to print its version upon startup (#93)
+- Authenticator client prints its version upon startup (#93)
```
sigalsax, line 11: Do we want to start with a verb instead of a noun to be in the same format as the entry below? So maybe "Add ability..."  ?

orenbm: but we don't "add an ability". this sounds like the client has an ability to print or not to print.

izgeri: also, for what it's worth, the TW have advised that since the header says  "Added" starting every line with "Add(s)" (or similar) is redundant and repetitive.

PR #123, opened 2020-04-22
Trigger a `kubernetes-conjur-demo` build after `master` build success

orenbm:
We would like to get an immediate feedback on a failure of a `master`
build of the project. As we don't have integration tests that run as
part of the build, our current way to get the feedback is by running
`kubernetes-conjur-demo` with the latest build.

By triggering `kubernetes-conjur-demo` to run after a successful
`master` build of this project it will be easier to pin point to a
change in case of a failure.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -59,6 +59,13 @@ pipeline {
   }
 
   post {
+    success {
+      script {
+        if (env.BRANCH_NAME == 'master') {
+          build (job:'conjurdemos--kubernetes-conjur-demo/master', wait: false)
+        }
+      }
+    }
     always {
       cleanupAndNotify(currentBuild.currentResult)
     }


izgeri:
the k8s-demo run uses the dockerhub images, not the build from the master branch of this project. this change will not have the desired effect. I did run k8s-demo using the master branch locally before adding the tag, and it ran successfully (as indicated on my version bump PR)

I recommend we close this issue, unless we add a build variable to k8s-demo that can set `LOCAL_AUTHENTICATOR=true` and build the authn-k8s and secretless projects before running the test cases (which will be expensive, and only increase the time it takes to run those tests)

orenbm:
Oh that's correct. I thought that we publish a docker image in every master build but that's incorrect.

Closing this issue.

Issue #126, opened 2020-05-01
Fix printing of user to logs


sgnn7:
## Summary

Changing the user from a simple string to a struct broke formatting for all logging of this field in both this project and others that use it.

## Steps to Reproduce
Use the project normally and notice the following in logs:

## Expected Results

```
INFO: 2020/04/20 08:23:44 authenticator.go:113: CAKC007I Logging in as user host/conjur/authn-k8s/conjur-5-e799170e-f-test/apps/test-app-5-e799170e-f/service_account/oc-test-app-summon-init.
```

## Actual Results (including error logs, if applicable)

```
INFO: 2020/04/20 08:23:44 authenticator.go:113: CAKC007I Logging in as user &{host/conjur/authn-k8s/conjur-5-e799170e-f-test/apps/test-app-5-e799170e-f/service_account/oc-test-app-summon-init host.conjur.authn-k8s.conjur-5-e799170e-f-test.apps test-app-5-e799170e-f.service_account.oc-test-app-summon-init}.
```

szh:
This is no longer broken:
https://github.com/cyberark/conjur-authn-k8s-client/blob/c02cf0d4fc21a4bbe6245ae51c09de70bc202fba/pkg/authenticator/k8s/authenticator.go#L158

PR #127, opened 2020-05-01
126 Fix username formatting

sgnn7:
### What does this PR do (include background context, if relevant)?

When Username changed from a simple string to a struct, none of its
references changed and all output showed the raw struct. Since `Suffix`
and `Prefix` fields are redundant, the new string formatter for this
object now prints the `FullUsername` field only which should fix both this
repo's output and any other repo that uses this module.

New output:
```
...
INFO: 2020/05/01 14:40:55 main.go:45: CAKC006I Authenticating as user 'host/path/to/policy/namespace/resource_type/resource_id'
...
INFO: 2020/05/01 14:40:55 authenticator.go:113: CAKC007I Logging in as user 'host/path/to/policy/namespace/resource_type/resource_id'
```

Connected to #126 

#### Change log
- [x] The CHANGELOG has been updated, or

#### Test coverage
- [x] This PR includes new unit and integration tests to go with the code changes, or

#### Follow-on issues
- [ ] Follow-up issue(s) have been logged (and links included below) to update documentation or related projects, or
- [x] No follow-up issues are required

#### Manual tests


pkg/authenticator/config/username_test.go
```diff
@@ -1,6 +1,8 @@
 package config
 
 import (
+	"fmt"
+
```
diverdane, line 5: Nit, maybe even not a problem... Do we need the space between the import of "fmt" and "testing"? They're both standard golang packages.

sgnn7: Blah I merged before seeing this - you're right and I'll keep this in mind for future changes on this repo!

PR #133, opened 2020-06-18
Add logs around client certificate usage

orenbm:
Connected to cyberark/conjur#1633

We now log that we loaded the certificate successfully and specify the
path that cert file was in. We also log that we deleted the certificate
so the user doesn't see that it's not there and assumes it wasn't copied
successfully from the Conjur server


orenbm:
@shulifink can you review the logs please?

pkg/authenticator/authenticator.go
```diff
@@ -142,6 +142,7 @@ func (auth *Authenticator) Login() error {
 
 		return log.RecordedError(log.CAKC012E, err.Error())
 	}
+	log.InfoLogger.Printf(log.CAKC015I, auth.Config.ClientCertPath)
```
Tovli, line 145: Are we ok with this, securitywise?
Maybe we can put just the last folder?

I would also consider using this only in DEBUG mode. Doesn't feel like a business information but something you would want to see during deployment only

pkg/authenticator/authenticator.go
```diff
@@ -153,6 +154,7 @@ func (auth *Authenticator) Login() error {
 
 	// clean up the client cert so it's only available in memory
 	os.Remove(auth.Config.ClientCertPath)
+	log.InfoLogger.Printf(log.CAKC016I)
```
Tovli, line 157: Securitywise - Can't this help attacker to understand the timeline he can get a memory dump with the cert?

pkg/log/log_messages.go
```diff
@@ -61,3 +61,5 @@ const CAKC011I string = "CAKC011I Login request to: %s"
 const CAKC012I string = "CAKC012I Authn request to: %s"
 const CAKC013I string = "CAKC013I Waiting for %s to re-authenticate"
 const CAKC014I string = "CAKC014I Kubernetes Authenticator Client v%s starting up..."
+const CAKC015I string = "CAKC015I Loaded client certificate successfully from %s"
```
shulifink, line 64: approved

pkg/log/log_messages.go
```diff
@@ -61,3 +61,5 @@ const CAKC011I string = "CAKC011I Login request to: %s"
...
+const CAKC016I string = "CAKC016I Deleted client certificate from memory"
```
shulifink, line 65: approved

pkg/authenticator/authenticator.go
shulifink: //Clean (capital C)

shulifink: //Remove the client cert.....

pkg/authenticator/authenticator.go
orenbm: 1. this field is configured by the user but i guess we can get almost the same value without printing the path and just printing that we loaded the certificate successfully. wdyt?
2. we don't have a debug log in the authn-client atm so i don't want to have a whole effort of adding such a logger, an env var, etc. We don't have many logs so it doesn't add a lot of noise.

pkg/authenticator/authenticator.go
orenbm: @shulifink thanks, but that is just a comment for developers. I am looking for feedback on the log messages.

orenbm: @shaharglazner do you see a security concern here? UX-wise there is value in this log message.

pkg/authenticator/authenticator.go
Tovli: 1 & 2: LGTM thanks

izgeri: Out of curiosity, wdym it's configurable? We have an open issue (afaik) in conjur to make it configurable, but last I looked its hardcoded in the server

izgeri: Also, I wish we had a debug log or configurable levels. These log lines are already really noisy in secretless and this PR is adding multiple new lines that will repeat as the authn cycle continues.

orenbm: > Out of curiosity, wdym it's configurable? We have an open issue (afaik) in conjur to make it configurable, but last I looked its hardcoded in the server

@izgeri you are correct. it is configurable only in the client (so it doesn't give any value). The point was that it is not a sensitive data as it is configurable in the env.

orenbm: > Also, I wish we had a debug log or configurable levels. These log lines are already really noisy in secretless and this PR is adding multiple new lines that will repeat as the authn cycle continues.

we actually have that for the secrets provider and can get it done here as well. I wouldn't have it in the current effort but I'll create an issue for it.

PR #145, opened 2020-08-11
Preparation for Secrets Provider deployment as a Job

sigalsax:
The Secrets Provider imports the authn-client as a package. Because we are working to support a Job, we need to distinguish between init container and Job deployments. To do, we need to add `SUPPORT_MULTIPLE_APPLICATIONS` env variable that when true represents Job deployment

This PR does the following:
- Removes CONTAINER_MODE from initial defaultConfig struct because we now support either ContainerMode="init" (init flow) or SupportMultipleApplications="true" (Job flow)
- Supports fetching SupportMultipleApplications value if it exists in environment

### DIFF
--- a/pkg/authenticator/config/config.go
+++ b/pkg/authenticator/config/config.go
@@ -12,25 +12,27 @@ import (
 // Config defines the configuration parameters
 // for the authentication requests
 type Config struct {
-	Account             string
-	ClientCertPath      string
-	ContainerMode       string
-	ConjurVersion       string
-	PodName             string
-	PodNamespace        string
-	SSLCertificate      []byte
-	TokenFilePath       string
-	TokenRefreshTimeout time.Duration
-	URL                 string
-	Username            *Username
+	Account                     string
+	ClientCertPath              string
+	ContainerMode               string
+	SupportMultipleApplications string
+	ConjurVersion               string
+	PodName                     string
+	PodNamespace                string
+	SSLCertificate              []byte
+	TokenFilePath               string
+	TokenRefreshTimeout         time.Duration
+	URL                         string
+	Username                    *Username
 }
 
 // Default settings (this comment added to satisfy linter)
 const (
-	DefaultClientCertPath = "/etc/conjur/ssl/client.pem"
-	DefaultTokenFilePath  = "/run/conjur/access-token"
-
-	DefaultConjurVersion = "5"
+	DefaultClientCertPath              = "/etc/conjur/ssl/client.pem"
+	DefaultTokenFilePath               = "/run/conjur/access-token"
+	DefaultContainerMode               = "sidecar"
+	DefaultSupportMultipleApplications = "false"
+	DefaultConjurVersion               = "5"
 
 	// DefaultTokenRefreshTimeout is the default time the system waits to reauthenticate on error
 	DefaultTokenRefreshTimeout = 6 * time.Minute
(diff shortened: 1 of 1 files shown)


pkg/authenticator/config/config.go
```diff
@@ -12,25 +12,27 @@ import (
...
-	PodNamespace        string
-	SSLCertificate      []byte
-	TokenFilePath       string
-	TokenRefreshTimeout time.Duration
-	URL                 string
-	Username            *Username
+	Account                     string
+	ClientCertPath              string
+	ContainerMode               string
+	SupportMultipleApplications string
```
izgeri, line 18: the rest of the params are alphabetical, can we maintain that?

pkg/authenticator/config/config.go
```diff
@@ -99,11 +101,10 @@ func populateConfig() (*Config, error) {
 	}
 
 	defaultConfig := &Config{
```
izgeri, line 103: I know you didn't change this, but in light of your other changes it seems odd to me that we call this var `defaultConfig` when it literally is instantiated with the custom config the user might set in the env

pkg/authenticator/config/config.go
```diff
@@ -143,5 +144,17 @@ func populateConfig() (*Config, error) {
...
+	defaultConfig.ContainerMode = DefaultContainerMode
+	// If CONTAINER_MODE is defined in the env we take its value
+	if envVal := os.Getenv("CONTAINER_MODE"); envVal != "" {
+		defaultConfig.ContainerMode = os.Getenv("CONTAINER_MODE")
+	}
+
+	defaultConfig.SupportMultipleApplications = DefaultSupportMultipleApplications
+	// If SUPPORT_MULTIPLE_APPS is defined in the env we take its value
+	if envVal := os.Getenv("SUPPORT_MULTIPLE_APPS"); envVal != "" {
+		defaultConfig.SupportMultipleApplications = os.Getenv("SUPPORT_MULTIPLE_APPS")
```
izgeri, line 156: in your PR description you say we support `CONTAINER_MODE` OR `SUPPORT_MULTIPLE_APPS`, but there is no logic here to only allow one or the other. does this code represent the flow we'd want?

izgeri:
also, is there a connected issue for this change?

pkg/authenticator/config/config.go
orenbm: @sigalsax i don't really understand the purpose of this new param. how will it be used? can we maybe leverage the existing param `CONTAINER_MODE` for this purpose?

for example, let the customer set `CONTAINER_MODE=job` and query for that (i know that `job` is not a container mode).

orenbm: if that's the issue then we can add a param `DEPLOYMENT_MODE` and use that instead of `SUPPORT_MULTIPLE_APPLICATIONS` to be more generic.

sigalsax: `DEPLOYMENT_MODE=job` is not sustainable or scalable because what happens when there's rotation, we are no longer a job. We are a deployment. It just gets a bit more confusing because what happens then if we run as a deployment but have another feature `DEPLOYMENT_MODE_deployment` will no longer be specific enough

We should start support flows instead of deployment types

sigalsax: When we support rotation (later milestone), we will add
```
SUPPORT_MULTIPLE_APPLICATIONS=true
ROTATION=true
```
That way the flows are clear

orenbm: you still didn't answer my question - how will `SUPPORT_MULTIPLE_APPLICATIONS` be used? what will we do if it's `true`? what will we do if it's `false`?

I'm trying to understand your intention to give a better review.

sigalsax: `SUPPORT_MULTIPLE_APPLICATIONS=true` means we will run as a job, if false we won't run as a Job and we expect to see `CONTAINER_MODE=init`
Because we are deploying with Helm, at this stage this env is not configurable. It will strictly be
```
- name: SUPPORT_MULTIPLE_APPS
   value: true
```
Usually for env var, we accept user input to fill in env var based on that input. For this env var, users will not be able to manipulate it. 
See this PR: https://github.com/cyberark/secrets-provider-for-k8s/pull/179

izgeri:
@sigalsax is this PR going to move forward, or should we close it for now?

sigalsax:
No moved the logic over the CSPFK

Issue #119, opened 2020-04-20
Errors are unnecessarily logged on client prior to certificate being injected


eranha:
Afte the CSR is sent the server and the response is being sent out-of-band, the client yields errors to log while attempting to load the certificate, until the server injects the certificate to the client. This type of output can be alarming to see and unhelpful so it would be good to only show this when there is a real problem getting the certificate.

AC:
- [x] Certificate reading errors should only be logged when we are sure that the server failed to inject the signed cert

```syslog
INFO: 2020/04/20 08:23:44 main.go:43: CAKC006I Authenticating as user '&{host/conjur/authn-k8s/conjur-5-e799170e-f-test/apps/test-app-5-e799170e-f/service_account/oc-test-app-summon-init host.conjur.authn-k8s.conjur-5-e799170e-f-test.apps test-app-5-e799170e-f.service_account.oc-test-app-summon-init}'
INFO: 2020/04/20 08:23:44 authenticator.go:181: CAKC005I Trying to login Conjur...
INFO: 2020/04/20 08:23:44 authenticator.go:113: CAKC007I Logging in as user &{host/conjur/authn-k8s/conjur-5-e799170e-f-test/apps/test-app-5-e799170e-f/service_account/oc-test-app-summon-init host.conjur.authn-k8s.conjur-5-e799170e-f-test.apps test-app-5-e799170e-f.service_account.oc-test-app-summon-init}.
INFO: 2020/04/20 08:23:44 requests.go:23: CAKC011I Login request to: https://conjur-follower.conjur-5-e799170e-f-test.svc.cluster.local/api/authn-k8s/conjur-5-e799170e-f-test/inject_client_cert
ERROR: 2020/04/20 08:23:44 authenticator.go:140: CAKC011E Client certificate not found at '/etc/conjur/ssl/client.pem'
ERROR: 2020/04/20 08:23:44 authenticator.go:184: CAKC015E Login failed
ERROR: 2020/04/20 08:23:44 main.go:46: CAKC016E Failed to authenticate
INFO: 2020/04/20 08:23:46 main.go:43: CAKC006I Authenticating as user '&{host/conjur/authn-k8s/conjur-5-e799170e-f-test/apps/test-app-5-e799170e-f/service_account/oc-test-app-summon-init host.conjur.authn-k8s.conjur-5-e799170e-f-test.apps test-app-5-e799170e-f.service_account.oc-test-app-summon-init}'
[...]

PR #153, opened 2020-08-28
Add check for client certificate existence with a timeout

micahlee:
### What does this PR do?
- _What's changed? Why were these changes made?_

This PR adds a wait for the client certificate file after login with a timeout. This provides some tolerance for delays between login and the client certificate becoming available to read in the container.

- _How should the reviewer approach this PR, especially if manual tests are required?_
- _Are there relevant screenshots you can add to the PR description?_

Connected to https://github.com/cyberark/conjur-authn-k8s-client/issues/119

#### Change log

#### Test coverage

#### Documentation

#### Manual tests


CHANGELOG.md
```diff
@@ -9,6 +9,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Fixed
 - Username formatting now correctly only prints FullUsername field ([#126](https://github.com/cyberark/conjur-authn-k8s-client/issues/126))
 
+### Changed
+- Wait slightly for the client certificate file to exist after login before
+  raising an error.
```
izgeri, line 14: this could be linked to #119

pkg/authenticator/config/config.go
```diff
@@ -143,5 +148,17 @@ func populateConfig() (*Config, error) {
...
+	defaultConfig.ClientCertTimeout = DefaultClientCertTimeout
+	clientCertTimeoutString := os.Getenv("CONJUR_CLIENT_CERT_TIMEOUT")
+	if len(clientCertTimeoutString) > 0 {
+		parsedClientCertTimeout, err := time.ParseDuration(clientCertTimeoutString)
+		if err != nil {
+			return nil, log.RecordedError(log.CAKC034E, err.Error())
+		}
+
+		defaultConfig.ClientCertTimeout = parsedClientCertTimeout
+	}
```
izgeri, line 161: we'd want to add some unit tests to this before it lands on master, if we end up merging this

pkg/authenticator/authenticator.go
```diff
@@ -133,6 +135,13 @@ func (auth *Authenticator) Login() error {
 		return log.RecordedError(log.CAKC029E, err.Error())
 	}
 
+	// ensure client certificate exists before attempting to read it, with a tolerance
+	// for small delays
+	err = waitForCertificate(auth.Config.ClientCertPath, auth.Config.ClientCertTimeout)
```
sgnn7, line 140: I'd probably name this a bit more specific (eg. `waitForInjectedCertificateFile`) or something similar

pkg/authenticator/authenticator.go
```diff
@@ -254,6 +263,25 @@ func (auth *Authenticator) ParseAuthenticationResponse(response []byte) error {
 	return nil
 }
 
+func waitForCertificate(certificatePath string, timeout time.Duration) error {
+	deadline := time.Now().Add(timeout)
```
sgnn7, line 267: We already use the [`backoff` library]("github.com/cenkalti/backoff") in the [main loop](https://github.com/cyberark/conjur-authn-k8s-client/blob/master/cmd/authenticator/main.go#L35-L43). You might as well use the same lib for the logic.

orenbm:
Thanks @micahlee !

i believe that the real fix for this should go in the server as, [at least from reading the code](https://github.com/cyberark/conjur/blob/a9f157ed526bc911c49271f1608e065a3b5e832e/app/domain/authentication/authn_k8s/inject_client_cert.rb#L91), it seems that the server waits for the server to be injected before sending a 200 OK response to the "inject_client_cert" request.

So we should fix the `validate_cert_installation` method so it really validates the cert is installed. 

we can still merge this as a workaround until the fix is in the server though, but we should not stop there.

pkg/authenticator/authenticator.go
```diff
@@ -133,6 +135,13 @@ func (auth *Authenticator) Login() error {
 		return log.RecordedError(log.CAKC029E, err.Error())
 	}
 
+	// ensure client certificate exists before attempting to read it, with a tolerance
+	// for small delays
+	err = waitForCertificate(auth.Config.ClientCertPath, auth.Config.ClientCertTimeout)
+	if err != nil {
+		return err
```
orenbm, line 142: do we want to return a `RecordedError` and print a log message?

pkg/authenticator/authenticator.go
```diff
@@ -254,6 +263,30 @@ func (auth *Authenticator) ParseAuthenticationResponse(response []byte) error {
 	return nil
 }
 
+func waitForCertificate(certificatePath string, timeout time.Duration) error {
```
orenbm, line 266: can we add a comment for this method as well if you did it for the rest?

pkg/authenticator/authenticator.go
```diff
@@ -254,6 +263,30 @@ func (auth *Authenticator) ParseAuthenticationResponse(response []byte) error {
...
+			// Certificate file is available
+			log.InfoLogger.Printf(log.CAKC017I, certificatePath)
+
+			// No error, the certificate exists within the deadline
+			return nil
+		}
+
+		if time.Now().After(deadline) {
+			// Timed out waiting for the certificate to exist
+			return log.RecordedError(log.CAKC033E, certificatePath)
```
orenbm, line 281: ah, i see see here that you recorded it. never mind my previous comment on this.

micahlee:
Thanks @orenbm!

> i believe that the real fix for this should go in the server as, at least from reading the code, it seems that the server waits for the server to be injected before sending a 200 OK response to the "inject_client_cert" request.

This is interesting, because this is also how I read that file, but I was advised in slack ([here](https://conjurhq.slack.com/archives/CGTRB6G8G/p1598624487211700?thread_ts=1598622442.205800&cid=CGTRB6G8G) )that this is not the case, and that the process is more async than that. It sounds like we may need to spend a little time verifying both what we expect of `/inject_client_cert`, and that it does what we expect.

In any case, yes, I agree that having the server verify successful injection would be better generally, rather than expecting the client to do this check. If we can do that, this should only need to be a temporary fix to mitigate the issue in the meantime.

orenbm:
[This PR](https://github.com/cyberark/conjur-authn-k8s-client/pull/155) might help as well. we can cherry pick its commits in here to have a merged image

pkg/authenticator/authenticator.go
rafis3: I know it's a little edge case, but since we only care about calculating elapsed time, can we use UTC so that it will not be affected by daylight saving time? I want to avoid unexpected behavior that could occur for 2 hours a year.

rafis3 (commented):
Thanks for this great fix @micahlee. Added one small comment as well.

pkg/authenticator/authenticator.go
```diff
@@ -133,6 +136,13 @@ func (auth *Authenticator) Login() error {
 		return log.RecordedError(log.CAKC029E, err.Error())
 	}
 
+	// ensure client certificate exists before attempting to read it, with a tolerance
```
sgnn7, line 139: nit: Comments should start with a capital letter

pkg/authenticator/authenticator.go
```diff
@@ -133,6 +136,13 @@ func (auth *Authenticator) Login() error {
...
+	// for small delays
+	err = auth.WaitForInjectedCertificateFile(auth.Config.ClientCertPath, auth.Config.ClientCertRetryCountLimit, os.Stat, os.IsNotExist)
```
sgnn7, line 141: Line-too-long

pkg/authenticator/authenticator.go
```diff
@@ -254,6 +264,35 @@ func (auth *Authenticator) ParseAuthenticationResponse(response []byte) error {
 	return nil
 }
 
+// We define these functions so we can mock WaitForInjectedCertificateFile
+type OsStatFunc func(name string) (os.FileInfo, error)
+type OsIsNotExistFunc func(err error) bool
```
sgnn7, line 269: Types should be placed before any functions in the file

pkg/authenticator/authenticator.go
```diff
@@ -254,6 +264,35 @@ func (auth *Authenticator) ParseAuthenticationResponse(response []byte) error {
...
+
+// WaitForInjectedCertificateFile waits for retryCountLimit seconds to see if the cert file
+// exists in the given path. If it's not there by the end of the retry count limit, it returns
+// an error.
+func (auth *Authenticator) WaitForInjectedCertificateFile(certificatePath string, retryCountLimit int, osStatFunc OsStatFunc, osIsNotExistFunc OsIsNotExistFunc) error {
```
sgnn7, line 274: As was briefly discussed in slack, this probably seems appropriate to extract to its own file

sgnn7: Line-too-long

pkg/authenticator/authenticator.go
```diff
@@ -254,6 +264,35 @@ func (auth *Authenticator) ParseAuthenticationResponse(response []byte) error {
...
+	constantBackOff := backoff.NewConstantBackOff(time.Second)
+	count := 0
+	return backoff.Retry(func() error {
+		info, err := osStatFunc(certificatePath)
+		if !osIsNotExistFunc(err) && !info.IsDir() {
```
sgnn7, line 279: I know this is old code but a check for "is it a file" seems more appropriate than "is it not a dir": https://golang.org/pkg/os/#FileMode.IsRegular

pkg/authenticator/authenticator_test.go
```diff
@@ -53,6 +55,28 @@ func TestAuthenticator(t *testing.T) {
 		})
 	})
 
+	Convey("WaitForInjectedCertificateFile", t, func() {
+		retryCountLimit := 10
+		authn := Authenticator{}
+		Convey("Returns nil if cert is installed", func() {
+			certificatePath := "exist"
+
+			So(authn.WaitForInjectedCertificateFile(certificatePath, retryCountLimit, mocks.MockOsStatFunc, mocks.MockOsIsNotExistFunc), ShouldBeNil)
```
sgnn7, line 64: Line-too-long

pkg/authenticator/authenticator_test.go
```diff
@@ -53,6 +55,28 @@ func TestAuthenticator(t *testing.T) {
...
+		Convey("Waits for whole time if cert is not installed", func() {
+			certificatePath := "path/to/non-existing/file"
+
+			expectedOutput := fmt.Errorf(
+				"CAKC033E Timed out after waiting for %d seconds for certificate file to exist: %s",
+				retryCountLimit,
+				certificatePath,
+			)
+
+			So(authn.WaitForInjectedCertificateFile(certificatePath, retryCountLimit, mocks.MockOsStatFunc, mocks.MockOsIsNotExistFunc), ShouldResemble, expectedOutput)
```
sgnn7, line 76: Line-too-long

pkg/authenticator/config/config.go
```diff
@@ -143,5 +149,17 @@ func populateConfig() (*Config, error) {
...
+	defaultConfig.ClientCertRetryCountLimit = DefaultClientCertRetryCountLimit
+	clientCertRetryCountLimitString := os.Getenv("CONJUR_CLIENT_CERT_RETRY_COUNT_LIMIT")
+	if len(clientCertRetryCountLimitString) > 0 {
+		parsedClientCertRetryCountLimit, err := strconv.Atoi(clientCertRetryCountLimitString)
+		if err != nil {
+			return nil, log.RecordedError(log.CAKC034E, err.Error())
+		}
+
+		defaultConfig.ClientCertRetryCountLimit = parsedClientCertRetryCountLimit
+	}
```
sgnn7, line 162: Not a blocker since it's old code but I think we've reached a point where some of this env var parsing could be done via helper methods

pkg/authenticator/mocks/os.go
```diff
@@ -0,0 +1,18 @@
...
+	"errors"
+	"os"
+)
+
+func MockOsStatFunc(name string) (os.FileInfo, error) {
+	if name == "exist" {
+		return &MockFileInfo{}, nil
+	} else {
+		return nil, errors.New("NotExist")
+	}
```
sgnn7, line 13:
```suggestion
	if name == "exist" {
		return &MockFileInfo{}, nil
	} 
	
	return nil, errors.New("NotExist")
```

pkg/log/log_messages.go
```diff
@@ -63,3 +65,4 @@ const CAKC013I string = "CAKC013I Waiting for %s to re-authenticate"
 const CAKC014I string = "CAKC014I Kubernetes Authenticator Client v%s starting up..."
 const CAKC015I string = "CAKC015I Loaded client certificate successfully from %s"
 const CAKC016I string = "CAKC016I Deleted client certificate from memory"
+const CAKC018I string = "CAKC018I Waiting for client certificate to become available..."
```
sgnn7, line 68: Should this be `CAKC017I`?

sgnn7 (changes requested):
@orenbm / @micahlee Left some comments

pkg/authenticator/authenticator.go
orenbm: would you prefer it like this?
```
	err = auth.WaitForInjectedCertificateFile(
		auth.Config.ClientCertPath,
		auth.Config.ClientCertRetryCountLimit,
		os.Stat,
		os.IsNotExist)
```
i don't like when function signatures span on more than one line.

pkg/authenticator/authenticator.go
orenbm: why would you move this to another file? would it be a simple util?

pkg/authenticator/authenticator.go
sgnn7: > i don't like when function signatures span on more than one line.

Me neither but long lines make maintenance much harder.

I don't remember well but I think this is what we used for the hanging lines:
```go
err = auth.WaitForInjectedCertificateFile(
    auth.Config.ClientCertPath,
    auth.Config.ClientCertRetryCountLimit,
    os.Stat,
    os.IsNotExist,
 )
```

and https://github.com/cyberark/secretless-broker/blob/master/internal/plugin/connectors/http/generic/connector.go#L62-L65 for the definitions

orenbm: ok this is actually the call but same question for the function signature

pkg/authenticator/authenticator.go
sgnn7: yeah simple util. the problem is that this file is after this PR at 350+ lines and getting hard to untangle.

pkg/authenticator/config/config.go
orenbm: thanks, leaving this out from this PR. i guess this helper should be used in secretless and secrets-provider as well.

pkg/authenticator/authenticator.go
izgeri: do we understand well how adding this wait will interact with the exponential backoff in main.go?

pkg/authenticator/config/config.go
```diff
@@ -143,5 +149,17 @@ func populateConfig() (*Config, error) {
 		defaultConfig.ClientCertPath = envVal
 	}
 
+	// Parse client cert retry count limit if one is provided from env
```
izgeri, line 152: should we have a test for this in config_test?

pkg/authenticator/authenticator.go
```diff
@@ -263,29 +264,29 @@ func (auth *Authenticator) ParseAuthenticationResponse(response []byte) error {
...
 			// No error, the certificate exists within the deadline
 			return nil
 		}
 
-		if time.Now().After(deadline) {
-			// Timed out waiting for the certificate to exist
-			return log.RecordedError(log.CAKC033E, certificatePath)
+		count++
+		if count >= retryCountLimit {
+			return backoff.Permanent(log.RecordedError(log.CAKC033E, retryCountLimit, certificatePath))
```
micahlee, line 286: I'm not super familiar with this backoff library, but I do see it has a facility for maximum retry attempts: https://godoc.org/github.com/cenkalti/backoff#WithMaxRetries

Did you look at this and determine it didn't do what you want? Or is this something to consider rather than keeping our own count?

pkg/authenticator/authenticator.go
orenbm: thanks, will do.

pkg/log/log_messages.go
orenbm: nice catch. i added 2 and removed one.

pkg/authenticator/config/config.go
orenbm: i looked at this a bit. we have this logic also in another instance but a bit different. i don't want to waste much time on this so i'll leave it out of scope.

pkg/authenticator/authenticator.go
orenbm: i actually moved to another backoff that @abrahamko implemented. it will be in `utils`.

pkg/authenticator/authenticator.go
orenbm: why will it matter? the exponential backoff has some calculated interval between each try, no matter how long that try had taken. am i wrong? do you have other concerns?

pkg/utils/wait_for_file_to_exist.go
```diff
@@ -0,0 +1,54 @@
...
+// WaitForFileToExist waits for retryCountLimit seconds to see if the file
+// exists in the given path. If it's not there by the end of the retry count limit, it returns
+// an error.
+func WaitForFileToExist(
+	path string,
+	retryCountLimit int,
+	verifyFileExistsFunc VerifyFileExistsFunc,
+) error {
+	if verifyFileExistsFunc == nil {
+		verifyFileExistsFunc = verifyFileExists
```
sgnn7, line 23: Not a blocker but to make it obvious that the right-hand is a function within this module rather than a variable, you can do something like:
```go
// Top of file
var DefaultVerifyFileExistsFunc = verifyFileExists
...
func Foo() {
    if verifyFileExistsFunc == nil {
        verifyFileExistsFunc = DefaultVerifyFileExistsFunc
    }
}
//
```

pkg/utils/limited_backoff_test.go
```diff
@@ -0,0 +1,75 @@
...
+
+				assertRetryCount(backOff, 0)
+
+				testLimitedBackOff(backOff, retryLimit, interval)
+			})
+		})
+	})
+}
+
+func testLimitedBackOff(backOff *limitedBackOff, retryLimit int, interval time.Duration) {
```
sgnn7, line 40: Not a blocker at all but this seems just a touch more complicated than it needs to be for the class you're testing.

sgnn7 (changes requested):
@orenbm Just a few more tiny comments

pkg/utils/wait_for_file_to_exist.go
orenbm: does it need to be capitalized? 

also - would it be decalred before or after `type VerifyFileExistsFunc func(path string) error`?

sgnn7: > does it need to be capitalized?

No but it makes it clear that it's a constant (even if in Golang it means that it's exported) so it's up to you.

> would it be decalred before or after type VerifyFileExistsFunc func(path string) error?

probably before

orenbm: done.

pkg/authenticator/config/config.go
izgeri: this concerns me - the config.go is poorly unit tested, and it's not going to get better if we add new logic and don't add unit tests for it.

sgnn7 (approved):
@orenbm LGTM - just need to get squash-merged to make sure our commit history is clean on it.

PR #155, opened 2020-08-29
Load cert from file if it exists

orenbm:
### What does this PR do?
The Login method sends a login request to the server and then tries to
read the cert file from the path where it should be injected to by the
server.

However, in case it is still not there then we fail and retry to login
but it's possible that the file is already there from the previous
request. In this case, we don't need to send another login request to
the server and we can read it from the path.

This will reduce the load on the server as we will have less requests to
 it.

#### Change log
- [X] This PR does not include user-facing changes and doesn't require a CHANGELOG update

#### Test coverage
- [X] The changes in this PR do not require tests

#### Documentation
- [X] This PR does not require updating any documentation

#### Manual tests
- [X] Manually run [Kubernetes-Conjur demo](https://github.com/conjurdemos/kubernetes-conjur-demo) with a local authn-k8s client image build

### DIFF
--- a/design/templates/solution_design.md
+++ b/design/templates/solution_design.md
@@ -1,0 +1,71 @@
+# Solution Design - Template
+[//]: # (Change the title above from "Template" to your design's title) 
+
+## Table of Contents
+[//]: # (You can use this tool to generate a TOC - https://ecotrust-canada.github.io/markdown-toc/) 
+
+## Glossary
+[//]: # (Describe terms that will be used throughout the design)
+[//]: # (You can use this tool to generate a table - https://www.tablesgenerator.com/markdown_tables#)
+
+| **Term** | **Description** |
+|----------|-----------------|
+|          |                 |
+|          |                 |
+
+## Useful links
+[//]: # (Add links that may be useful for the reader)
+
+## Background
+[//]: # (Give relevant background for the designed feature. What is the motivation for this solution?)
+
+## Issue description
+[//]: # (Elaborate on the issue you are writing a solution for)
+
+## Solution
+[//]: # (Elaborate on the solution you are suggesting in this page. Address the functional requirements and the non functional requirements that this solution is addressing. If there are a few options considered for the solution, mention them and explain why the actual solution was chosen over them. Add an execution plan when relevant. It doesn't have to be a full breakdown of the feature, but just a recommendation to how the solution should be approached.)
+
+### Design
+[//]: # (Add any diagrams, charts and explanations about the design aspect of the solution. Elaborate also about the expected user experience for the feature)
+
+### Backwards compatibility
+[//]: # (Address how you are going to handle backwards compatibility, if necessary)
+
+### Performance
+[//]: # (Elaborate on whether your solution will affect the product's performance, and how)
+
+### Affected Components
+[//]: # (Address the components that will be affected by your solution [Conjur, DAP, clients, integrations, etc.])
+
+## Security
+[//]: # (Are there any security issues with your solution and how do you plan to mitigate them? Even if you mentioned them somewhere in the doc it may be convenient for the security architect review to have them centralized here)
+
+## Test Plan
+[//]: # (Fill in the table below to depict the tests that should run to validate your solution)
+[//]: # (You can use this tool to generate a table - https://www.tablesgenerator.com/markdown_tables#)
+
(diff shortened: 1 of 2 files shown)


pkg/authenticator/authenticator.go
```diff
@@ -112,28 +112,65 @@ func (auth *Authenticator) Login() error {
...
+	return auth.PublicCert != nil
+}
+
+// CertFileExist returns true if certificate file exists in the ClientCertPath
+func (auth *Authenticator) CertFileExist() bool {
+	info, err := os.Stat(auth.Config.ClientCertPath)
+	if os.IsNotExist(err) {
+		return false
 	}
+	return !info.IsDir()
```
sigalsax, line 158: this return seems a bit weird to me because in the above if statement, you are returning false if it was reported that a file or directory does not exist but then you are returning the certPath is not a file/directory with `!info.IsDir()` which is unreachable.

Shouldn't you return a condition that evaluates to true?

pkg/authenticator/authenticator.go
```diff
@@ -112,28 +112,65 @@ func (auth *Authenticator) Login() error {
...
+	return nil
+}
+
+// Returns true if we are logged in (have a cert)
+func (auth *Authenticator) IsLoggedIn() bool {
+	return auth.PublicCert != nil
+}
+
+// CertFileExist returns true if certificate file exists in the ClientCertPath
+func (auth *Authenticator) CertFileExist() bool {
```
sigalsax, line 153: Can we add unit tests for this?

design/templates/solution_design.md
```diff
@@ -0,0 +1,71 @@
+# Solution Design - Template
+[//]: # (Change the title above from "Template" to your design's title)
```
sigalsax, line 2: was is the template for the solution design added with this PR?

sigalsax (commented):
left some comments

design/templates/solution_design.md
izgeri: could be a sign that a rebase is needed?

pkg/authenticator/authenticator.go
```diff
@@ -112,6 +112,25 @@ func (auth *Authenticator) Login() error {
 
 	log.InfoLogger.Printf(log.CAKC007I, auth.Config.Username)
 
+	// The cert file may be present from a previous login request
+	if !auth.CertFileExist() {
```
izgeri, line 116: what if the cert file exists, but is no longer valid? don't we have a method to check if the cert is valid in one of these packages? should that be invoked here?

pkg/authenticator/authenticator.go
```diff
@@ -133,7 +152,35 @@ func (auth *Authenticator) Login() error {
 		return log.RecordedError(log.CAKC029E, err.Error())
 	}
 
-	// load client cert
+	return nil
+}
+
+// Returns true if we are logged in (have a cert)
+func (auth *Authenticator) IsLoggedIn() bool {
+	return auth.PublicCert != nil
```
izgeri, line 160: again, couldn't this present a problem if we have a cert but it is no longer valid?

pkg/authenticator/authenticator.go
```diff
@@ -133,7 +152,35 @@ func (auth *Authenticator) Login() error {
...
+}
+
+// Returns true if certificate is expired or close to expiring
+func (auth *Authenticator) IsCertExpired() bool {
+	certExpiresOn := auth.PublicCert.NotAfter.UTC()
+	currentDate := time.Now().UTC()
+
+	log.InfoLogger.Printf(log.CAKC008I, certExpiresOn)
+	log.InfoLogger.Printf(log.CAKC009I, currentDate)
+	log.InfoLogger.Printf(log.CAKC010I, bufferTime)
```
izgeri, line 176: why are these being logged? I already have concerns about the number of log entries in this project since the values are also logged in secretless and it creates a LOT of extra log messages - here we add three, that look more like they should be debug than info level

izgeri: edit to add: I see that this is how we did this below, but my concern about these being debug and not info messages stands

orenbm:
Now that we add a timeout for waiting to the client cert to exist in #153, we can close this PR. This is because if the cert is not installed in 10 seconds we assume that it won't be installed later so it doesn't make sense to check if it exists before sending another login request.

PR #160, opened 2020-09-10
Extract env var value read into a util

orenbm:
We would like to better UT the logic of reading a value from the env
and defaulting to a default value if it's not present. We also have the
logic that transforms this env var value into an int or a duration (and
maybe more in the future) and it is best that this logic is UTed and is
implemented in its own util. This improves the readability of the
config.go file and of the util functions themselves.

This will also let us use these utils in other projects that consume the
authn-client without implementing them there as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,9 @@ questions, please contact us on [Discourse](https://discuss.cyberarkcommons.org/
 - `CONJUR_AUTHN_URL`: URL pointing to authenticator service endpoint
 - `CONJUR_AUTHN_LOGIN`: Host login for pod e.g. `namespace/service_account/some_service_account`
 - `CONJUR_SSL_CERTIFICATE`: Public SSL cert for Conjur connection
-- `CONJUR_TOKEN_TIMEOUT`: Timeout for fetching a new token (defaults to 6 minutes). In most cases, this variable should not be modified.
+- `CONJUR_TOKEN_TIMEOUT`: Timeout for fetching a new token (defaults to 6 minutes). 
+                          In most cases, this variable should not be modified. The value should be in a
+                          format that can be parsed with [time.ParseDuration](https://golang.org/pkg/time/#ParseDuration) (e.g "6m0s")
 
 Flow:
 
--- a/pkg/authenticator/config/config.go
+++ b/pkg/authenticator/config/config.go
@@ -4,10 +4,10 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
-	"strconv"
 	"time"
 
 	"github.com/cyberark/conjur-authn-k8s-client/pkg/log"
+	"github.com/cyberark/conjur-authn-k8s-client/pkg/utils"
 )
 
 // Config defines the configuration parameters
@@ -35,11 +35,11 @@ const (
 	DefaultConjurVersion = "5"
 
 	// DefaultTokenRefreshTimeout is the default time the system waits to reauthenticate on error
-	DefaultTokenRefreshTimeout = 6 * time.Minute
+	DefaultTokenRefreshTimeout = "6m0s"
 
 	// DefaultClientCertRetryCountLimit is the amount of times we wait after successful
 	// login for the client certificate file to exist, where each time we wait for a second.
-	DefaultClientCertRetryCountLimit = 10
+	DefaultClientCertRetryCountLimit = "10"
 )
 
 var requiredEnvVariables = []string{
@@ -104,7 +104,7 @@ func populateConfig() (*Config, error) {
 		}
 	}
 
-	defaultConfig := &Config{
+	config := &Config{
 		Account:       os.Getenv("CONJUR_ACCOUNT"),
 		ContainerMode: os.Getenv("CONTAINER_MODE"),
 		PodName:       os.Getenv("MY_POD_NAME"),
@@ -113,10 +113,10 @@ func populateConfig() (*Config, error) {
 	}
 
 	// Only versions '4' & '5' are allowed, with '5' being used as the default
-	defaultConfig.ConjurVersion = DefaultConjurVersion
+	config.ConjurVersion = DefaultConjurVersion
 	switch os.Getenv("CONJUR_VERSION") {
 	case "4":
-		defaultConfig.ConjurVersion = "4"
+		config.ConjurVersion = "4"
 	case "5":
 		break // Stick with default
 	case "":
@@ -126,40 +126,38 @@ func populateConfig() (*Config, error) {
 	}
 
 	// Parse token refresh rate if one is provided from env
-	defaultConfig.TokenRefreshTimeout = DefaultTokenRefreshTimeout
-	tokenRefreshTimeoutString := os.Getenv("CONJUR_TOKEN_TIMEOUT")
-	if len(tokenRefreshTimeoutString) > 0 {
-		parsedTokenRefreshTimeout, err := time.ParseDuration(tokenRefreshTimeoutString)
-		if err != nil {
-			return nil, log.RecordedError(log.CAKC010E, err.Error())
-		}
-
-		defaultConfig.TokenRefreshTimeout = parsedTokenRefreshTimeout
+	tokenRefreshTimeout, err := utils.DurationFromEnvOrDefault(
+		"CONJUR_TOKEN_TIMEOUT",
+		DefaultTokenRefreshTimeout,
+		nil,
+	)
+	if err != nil {
+		return nil, err
 	}
+	config.TokenRefreshTimeout = tokenRefreshTimeout
 
-	defaultConfig.TokenFilePath = DefaultTokenFilePath
+	config.TokenFilePath = DefaultTokenFilePath
 	// If CONJUR_TOKEN_FILE_PATH is defined in the env we take its value
 	if envVal := os.Getenv("CONJUR_AUTHN_TOKEN_FILE"); envVal != "" {
-		defaultConfig.TokenFilePath = envVal
+		config.TokenFilePath = envVal
 	}
 
-	defaultConfig.ClientCertPath = DefaultClientCertPath
+	config.ClientCertPath = DefaultClientCertPath
 	// If CONJUR_CLIENT_CERT_PATH is defined in the env we take its value
 	if envVal := os.Getenv("CONJUR_CLIENT_CERT_PATH"); envVal != "" {
-		defaultConfig.ClientCertPath = envVal
+		config.ClientCertPath = envVal
 	}
 
 	// Parse client cert retry count limit if one is provided from env
-	defaultConfig.ClientCertRetryCountLimit = DefaultClientCertRetryCountLimit
-	clientCertRetryCountLimitString := os.Getenv("CONJUR_CLIENT_CERT_RETRY_COUNT_LIMIT")
-	if len(clientCertRetryCountLimitString) > 0 {
-		parsedClientCertRetryCountLimit, err := strconv.Atoi(clientCertRetryCountLimitString)
-		if err != nil {
-			return nil, log.RecordedError(log.CAKC034E, err.Error())
-		}
-
-		defaultConfig.ClientCertRetryCountLimit = parsedClientCertRetryCountLimit
+	clientCertRetryCountLimit, err := utils.IntFromEnvOrDefault(
+		"CONJUR_CLIENT_CERT_RETRY_COUNT_LIMIT",
+		DefaultClientCertRetryCountLimit,
+		nil,
+	)
+	if err != nil {
+		return nil, err
 	}
+	config.ClientCertRetryCountLimit = clientCertRetryCountLimit
 
-	return defaultConfig, nil
+	return config, nil
 }
--- a/pkg/utils/environment_variable.go
+++ b/pkg/utils/environment_variable.go
@@ -1,0 +1,61 @@
+package utils
+
+import (
+	"os"
+	"strconv"
+	"time"
+
+	"github.com/cyberark/conjur-authn-k8s-client/pkg/log"
+)
+
+// GetEnvFunc is a function that returns a value for an environment variable
+type GetEnvFunc func(envVarName string) string
+
+// defaultGetEnv is the default GetEnvFunc. It uses the standard OS Getenv.
+var defaultGetEnvFunc = os.Getenv
+
+// IntFromEnvOrDefault extracts the value of the given environment variable,
+// or the default value if the environment variable is not defined, and returns
+// it as an int object
+func IntFromEnvOrDefault(
+	envVarName string,
+	defaultValue string,
+	getEnv GetEnvFunc,
+) (int, error) {
+	value := envVarValueOrDefault(envVarName, defaultValue, getEnv)
+	valueInt, err := strconv.Atoi(value)
+	if err != nil {
+		return 0, log.RecordedError(log.CAKC010E, envVarName, err.Error())
+	}
+
+	return valueInt, nil
+}
+
+// DurationFromEnvOrDefault extracts the value of the given environment variable,
+// or the default value if the environment variable is not defined, and returns
+// it as a time.Duration object
+func DurationFromEnvOrDefault(
+	envVarName string,
+	defaultValue string,
+	getEnv GetEnvFunc,
+) (time.Duration, error) {
+	value := envVarValueOrDefault(envVarName, defaultValue, getEnv)
+	valueDuration, err := time.ParseDuration(value)
+	if err != nil {
+		return 0, log.RecordedError(log.CAKC010E, envVarName, err.Error())
+	}
+
+	return valueDuration, nil
+}
+
+func envVarValueOrDefault(envVarName string, defaultValue string, getEnv GetEnvFunc) string {
+	if getEnv == nil {
+		getEnv = defaultGetEnvFunc
+	}
+	value := getEnv(envVarName)
+	if len(value) > 0 {
+		return value
+	}
+
+	return defaultValue
+}
--- a/pkg/utils/environment_variable_test.go
+++ b/pkg/utils/environment_variable_test.go
@@ -1,0 +1,170 @@
+package utils
+
+import (
+	"fmt"
+	"testing"
+	"time"
+
+	. "github.com/smartystreets/goconvey/convey"
+)
+
+const (
+	envVarName              = "SOME_ENV_VAR"
+	intValueStr             = "20"
+	expectedIntValue        = 20
+	defaultIntValueStr      = "10"
+	defaultIntValue         = 10
+	durationValueStr        = "20m0s"
+	expectedDurationValue   = 20 * time.Minute
+	defaultDurationValueStr = "10m0s"
+	defaultDurationValue    = 10 * time.Minute
+	InvalidValueStr         = "SOME_ENV_VAR_VALUE"
+)
+
+// mockGetEnv structure is used to implement a getEnvFunc function that
+// returns a canned value that is set when the structure is initialized.
+type mockGetEnv struct {
+	value string
+}
+
+// newMockGetEnv creates and initializes a mockGetEnv structure
+func newMockGetEnv(value string) mockGetEnv {
+	return mockGetEnv{value}
+}
+
+// getEnv implements the getEnvFunc type for a mockGetEnv struct
+func (m mockGetEnv) getEnv(name string) string {
+	return m.value
+}
+
+func TestEnvironmentVariable(t *testing.T) {
+	Convey("IntFromEnvOrDefault", t, func() {
+		Convey("When parsing an int value", func() {
+			Convey("That was retrieved from the env", func() {
+				mockGetEnv := newMockGetEnv(intValueStr)
+				intValue, err := IntFromEnvOrDefault(
+					envVarName,
+					defaultIntValueStr,
+					mockGetEnv.getEnv,
+				)
+
+				Convey("Finishes without raising an error", func() {
+					So(err, ShouldEqual, nil)
+				})
+
+				Convey("Parses the value into an int", func() {
+					So(
+						intValue,
+						ShouldResemble,
+						expectedIntValue,
+					)
+				})
+			})
+
+			Convey("That was given as a default value", func() {
+				mockGetEnv := newMockGetEnv("")
+				intValue, err := IntFromEnvOrDefault(
+					envVarName,
+					defaultIntValueStr,
+					mockGetEnv.getEnv,
+				)
+
+				Convey("Finishes without raising an error", func() {
+					So(err, ShouldEqual, nil)
+				})
+
+				Convey("Parses the value into an int", func() {
+					So(
+						intValue,
+						ShouldResemble,
+						defaultIntValue,
+					)
+				})
+			})
+		})
+
+		Convey("When parsing a non-int value", func() {
+			mockGetEnv := newMockGetEnv(InvalidValueStr)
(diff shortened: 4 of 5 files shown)


pkg/authenticator/config/config.go
```diff
@@ -35,11 +35,11 @@ const (
 	DefaultConjurVersion = "5"
 
 	// DefaultTokenRefreshTimeout is the default time the system waits to reauthenticate on error
-	DefaultTokenRefreshTimeout = 6 * time.Minute
+	DefaultTokenRefreshTimeout = "6m0s"
```
sigalsax, line 38: in SP I wrote with `time` was there a particular reason why this was changed?

orenbm: the util receives a string as an input to match the value that will be given in the env. i think that it makes sense that both the value from the env and the default value are of the same type. after we get the value (from the env or default) we convert it to the required type (int, duration, etc.).

pkg/utils/environment_variable.go
```diff
@@ -0,0 +1,58 @@
...
+	envVarName string,
+	defaultValue string,
+	envVarValueOrDefaultFunc EnvVarValueOrDefaultFunc,
+) (time.Duration, error) {
+	if envVarValueOrDefaultFunc == nil {
+		envVarValueOrDefaultFunc = defaultEnvVarValueOrDefault
+	}
+
+	envVarValue := envVarValueOrDefault(envVarName, defaultValue)
+	envVarValueDuration, err := time.ParseDuration(envVarValue)
```
sigalsax, line 43: why is this needed when you can just get it via `6*time.seconds`?

orenbm: but what if the value is from the env and not the default value? in that case it will be a string.

orenbm: that's why i think it's more simple to have them both as strings. this will be the case for every env var regardless to the expected type. The value is always a string and the logic that converts it to the required type is in the util. it will create a uniformity throughout configurable values in this repo and in our other k8s projects.

pkg/utils/environment_variable.go
```diff
@@ -0,0 +1,58 @@
...
+var defaultEnvVarValueOrDefault = envVarValueOrDefault
+
+type EnvVarValueOrDefaultFunc func(envVarName string, defaultValue string) string
+
+func IntFromEnvOrDefault(
+	envVarName string,
+	defaultValue string,
+	envVarValueOrDefaultFunc EnvVarValueOrDefaultFunc,
+) (int, error) {
+	if envVarValueOrDefaultFunc == nil {
```
sigalsax, line 20: when would this not be `nil`? an example would help me understand what we are trying to accomplish here (:

sigalsax: IMO this breaking up of parameters style is really hard for my 👀  to follow

orenbm: look at the UT :)

orenbm: this is the way we can pass mocks to the functions in go

pkg/utils/environment_variable.go
```diff
@@ -0,0 +1,58 @@
...
+		envVarValueOrDefaultFunc = defaultEnvVarValueOrDefault
+	}
+
+	envVarValue := envVarValueOrDefault(envVarName, defaultValue)
+	envVarValueInt, err := strconv.Atoi(envVarValue)
+	if err != nil {
+		return -1, log.RecordedError(log.CAKC010E, envVarName, err.Error())
```
sigalsax, line 27: return -1 = error or is it return 1?

pkg/utils/environment_variable.go
orenbm: > IMO this breaking up of parameters style is really hard for my 👀 to follow

i understand. i had a conversation about this with @sgnn7 yesterday. the alternative is a line that is too long which is an issue as well.

pkg/utils/environment_variable.go
orenbm: `-1` is the way for me to pass `nil` for an int. it's irrelevant as if `err` is not nil we don't use the returned value.

pkg/authenticator/config/config.go
```diff
@@ -35,11 +35,11 @@ const (
...
 
 	// DefaultClientCertRetryCountLimit is the amount of times we wait after successful
 	// login for the client certificate file to exist, where each time we wait for a second.
-	DefaultClientCertRetryCountLimit = 10
+	DefaultClientCertRetryCountLimit = "10"
```
sigalsax, line 42: are there minimum or maximum values for this limit? Can they put `-5` for example? Would the logic hold up against negative numbers?

sigalsax: In SP we added a check for `envValueInt < minValue` with a minValue being 0 in order to avoid this

pkg/utils/environment_variable.go
```diff
@@ -0,0 +1,58 @@
+package utils
```
sigalsax, line 1: In terms of the naming of this package, it doesn't tell me much. So IMO add a comment at the top of the file what benefit this util gives us or change the name. I vote for comment (:

pkg/authenticator/config/config.go
orenbm: hmmmm. do you think it is needed or is it too low level?

orenbm: do we want to not let users set (for example) `CONJUR_TOKEN_TIMEOUT=-5`?

pkg/utils/environment_variable.go
orenbm: yeah the package name is for all the utils in this project. i will add a comment

orenbm: Thinking about it again, i think that the file `utils/environment_variable.go` is quite self-explanatory. would the comment `This file centralizes functions for extracting values from environment variables` help the reader understand it better? I think it does not.

orenbm: i added comments for each method instead

pkg/utils/environment_variable.go
sigalsax: the 6 would be from the env and *time*seconds would be from the code

orenbm: yuck (sorry)

orenbm: also - the env should be `6m0s` to match 6 minutes. just `6` won't work.

sigalsax: But no problem. You gave a good explanation

pkg/authenticator/config/config.go
sigalsax: I can totally see this being an issue _especially_ because it is configurable no? It is coming from an environment and I don't trust user input 🙃

pkg/utils/environment_variable.go
orenbm: would you prefer if the env var would be an int? e.g `CONJUR_TOKEN_TIMEOUT=6`? if so we would change the logic to `strconv.Atoi(envVarValue) * time.Minutes`

pkg/authenticator/config/config.go
orenbm: yes, but it adds another param that the function consumer needs to set where for most of the times it will be `0` and not interesting. i'll let @izgeri give her take on this before going forward.

pkg/utils/environment_variable.go
orenbm: the thing is that this will be a breaking change as today passing an int will not work. However, i believe that this variable was never used correctly and that setting `CONJUR_TOKEN_TIMEOUT=6` does not work also today. verifying now.

orenbm: yep, this won't work in the current product. i get a `ERROR: 2020/09/10 11:56:22 config.go:128: CAKC010E Failed to parse CONJUR_TOKEN_TIMEOUT. Reason: time: missing unit in duration 6` error.

@izgeri when this env var was created how did you intend to set it? with `6m0s`? do you know who defined this?
do you see a problem with changing this to an int input so it will be `CONJUR_TOKEN_TIMEOUT="6"` instead of `CONJUR_TOKEN_TIMEOUT="6m0s"`?

orenbm: once we get to a decision i will update the README with the valid value pattern.

pkg/authenticator/config/config.go
```diff
@@ -150,16 +149,15 @@ func populateConfig() (*Config, error) {
...
+		"CONJUR_CLIENT_CERT_RETRY_COUNT_LIMIT",
+		DefaultClientCertRetryCountLimit,
+		nil,
+	)
+	if err != nil {
+		return nil, err
 	}
+	defaultConfig.ClientCertRetryCountLimit = clientCertRetryCountLimit
 
 	return defaultConfig, nil
```
izgeri, line 162: if we are updating this, can we change the name of this var? it's _not_ the default config, it's the config that the client will use based on the default config _and_ the user-provided config

orenbm: sure, nice catch.

pkg/utils/environment_variable.go
orenbm: aha, i can see that this env var was used before as:
```
- name: CONJUR_TOKEN_TIMEOUT
+            value: "15s"
``` 
(so it in Slack) so i wouldn't change the UX to be of int to not break changes.

orenbm: @izgeri can you please review the commit i added with the readme change?

pkg/utils/environment_variable.go
```diff
@@ -0,0 +1,64 @@
...
+
+import (
+	"os"
+	"strconv"
+	"time"
+
+	"github.com/cyberark/conjur-authn-k8s-client/pkg/log"
+)
+
+var defaultEnvVarValueOrDefault = envVarValueOrDefault
```
izgeri, line 11: what is happening on this line?

orenbm: we set the function as a variable so it's clearer to the reader that the function being used in the one defined at the env of the file

CHANGELOG.md
```diff
@@ -8,6 +8,13 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Fixed
 - Username formatting now correctly only prints FullUsername field ([#126](https://github.com/cyberark/conjur-authn-k8s-client/issues/126))
+- Timeout for token retrievals is now correctly reset in consequent failures
+  ([cyberark/conjur-authn-k8s-client#158](https://github.com/cyberark/conjur-authn-k8s-client/issues/158))
```
izgeri, line 12:
```suggestion
- Username formatting now correctly only prints `FullUsername` field.
   [cyberark/conjur-authn-k8s-client#126](https://github.com/cyberark/conjur-authn-k8s-client/issues/126))
- Timeout for token retrievals is now correctly reset in consecutive authentication failures.
  [cyberark/conjur-authn-k8s-client#158](https://github.com/cyberark/conjur-authn-k8s-client/issues/158)
```

I am not sure the user impact from these changes is clear

for the first, is it that the log messages now print a more succinct name or a more detailed name? I'd call that out specifically
> Logs now correctly print only the Conjur identity without the policy branch prefix.

for the second option, it's more like
> When authentication fails, the exponential backoff retry is correctly reset so that it will continue to attempt to authenticate until backoff is exhausted.

CHANGELOG.md
```diff
@@ -8,6 +8,13 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
...
+
+### Changed
+- Wait slightly for the client certificate file to exist after login before
+  raising an error.
```
izgeri, line 16: What is the user impact of this change? How does it affect the UX?

README.md
```diff
@@ -38,7 +38,7 @@ questions, please contact us on [Discourse](https://discuss.cyberarkcommons.org/
 - `CONJUR_AUTHN_URL`: URL pointing to authenticator service endpoint
 - `CONJUR_AUTHN_LOGIN`: Host login for pod e.g. `namespace/service_account/some_service_account`
 - `CONJUR_SSL_CERTIFICATE`: Public SSL cert for Conjur connection
-- `CONJUR_TOKEN_TIMEOUT`: Timeout for fetching a new token (defaults to 6 minutes). In most cases, this variable should not be modified.
+- `CONJUR_TOKEN_TIMEOUT`: Timeout for fetching a new token (defaults to 6 minutes). In most cases, this variable should not be modified. The value should be in a signed sequence of decimal numbers with optional fraction and unit suffix (e.g "6m0s")
```
izgeri, line 41: this should be broken into multiple lines

> a signed sequence of decimal numbers with optional fraction and unit suffix
I don't know what this means without the example. is there a standard we could link to here?

design/templates/solution_design.md
```diff
@@ -0,0 +1,71 @@
+# Solution Design - Template
```
izgeri, line 1: does this indicate your branch may need a rebase?

pkg/authenticator/config/config.go
izgeri: I think we should have error checking to ensure sane and acceptable values are input for this param. is the param currently documented? in the README or just in the CONTRIB? if it's just for dev purposes / dev troubleshooting, we could just leave undocumented for now. if we're adding it AND documenting it, then we need to be more careful (and this would need a changelog entry)

CHANGELOG.md
orenbm: the client will fail less frequently as we now wait for up to 10 seconds to read the cert instead of immediately reading it. We are also logging our efforts so that's another customer-facing change. would you change the entry according to this?

pkg/utils/environment_variable.go
```diff
@@ -0,0 +1,64 @@
...
+
+import (
+	"os"
+	"strconv"
+	"time"
+
+	"github.com/cyberark/conjur-authn-k8s-client/pkg/log"
+)
+
+var defaultEnvVarValueOrDefaultFunc = envVarValueOrDefault
```
izgeri, line 11: what is happening on this line?

pkg/utils/environment_variable.go
izgeri: might be worth a comment - it's not a pattern I've seen before. maybe other devs would disagree but @sigalsax already had a question :)

pkg/utils/environment_variable.go
izgeri: I'm pretty sure that `CONJUR_TOKEN_TIMEOUT` was added for dev troubleshooting, though I see it is documented in the README

I think it is wise to set a standard and enforce it, we should just make sure that the standard (and the default, btw) is clear

izgeri (changes requested):
left a handful of comments, still would be good for @sgnn7 to weigh in since he chatted with you about some of the implementation details

README.md
orenbm: i took it from the golang docs for `time`: https://golang.org/pkg/time/#ParseDuration. i'll add a link

orenbm: how about `The value should be in a format that can be parsed with [time.ParseDuration](https://golang.org/pkg/time/#ParseDuration) (e.g "6m0s")`?

design/templates/solution_design.md
orenbm: correct

orenbm:
> I think we should have error checking to ensure sane and acceptable values are input for this param. is the param currently documented? in the README or just in the CONTRIB? if it's just for dev purposes / dev troubleshooting, we could just leave undocumented for now. if we're adding it AND documenting it, then we need to be more careful (and this would need a changelog entry)

`CONJUR_CLIENT_CERT_RETRY_COUNT_LIMIT` is for dev purposes so i'm leaving it undocumented. It is verified to be an int value when parsing it.

pkg/utils/environment_variable.go
orenbm: added a comment
```
// The function 'envVarValueOrDefault' is defined below. We set this var at the
// top of the file so when we use the function it is clear that it's the default one.
```

orenbm:
> might be worth a comment - it's not a pattern I've seen before. maybe other devs would disagree but @sigalsax already had a question :)

@izgeri added a comment

orenbm:
> I'm pretty sure that CONJUR_TOKEN_TIMEOUT was added for dev troubleshooting, though I see it is documented in the README

I think it is wise to set a standard and enforce it, we should just make sure that the standard (and the default, btw) is clear

we enforce it in `DurationFromEnvOrDefault`. an error will be raised if the value is not in the valid format.

pkg/utils/environment_variable_test.go
```diff
@@ -0,0 +1,105 @@
...
+	InvalidValueStr        = "SOME_ENV_VAR_VALUE"
+)
+
+// This mock function always returns the given default value
+func MockEnvVarValueOrDefaultFunc(envVarName string, defaultValue string) string {
+	return defaultValue
+}
+
+func TestEnvironmentVariable(t *testing.T) {
+	Convey("IntFromEnvOrDefault", t, func() {
```
diverdane, line 26: I think we're deprecating the use of `Convey`? But personally I'm fine with it... it tells the story of what you're testing.

pkg/authenticator/config/config.go
diverdane: I don't want to hold up this PR, but we should really consider doing a followup to allow for minimum/maximum range validation. In most cases, I'd bet that we can define hard-coded values for min and max values that are sensible and are liberal enough to cover all cases. For example, for certificate retries, we could set the minimum to 0 and the maximum to something astronomical (`2**32 - 1`?) If we don't add range checks, I don't think we know what would happen if someone sets retries or some other configured numbered value to -1 (I'm sure it's not in our integration testing).

diverdane (commented):
This is good stuff, thanks for moving this logic to a re-usable util package!
I've included some suggestions on how the code might be refactored to simplify and maybe help readability and code coverage.

pkg/utils/environment_variable.go
```diff
@@ -0,0 +1,68 @@
...
+	if envVarValueOrDefaultFunc == nil {
+		envVarValueOrDefaultFunc = defaultEnvVarValueOrDefaultFunc
+	}
+
+	envVarValue := envVarValueOrDefault(envVarName, defaultValue)
+	envVarValueInt, err := strconv.Atoi(envVarValue)
+	if err != nil {
+		// We return -1 as we can't return nil for int. This value will not be used
+		// as err is not nil.
+		return -1, log.RecordedError(log.CAKC010E, envVarName, err.Error())
```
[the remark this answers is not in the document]
orenbm: nice, great to learn this!

@sigalsax, @izgeri  ^

pkg/utils/environment_variable.go
```diff
@@ -0,0 +1,68 @@
...
+	"time"
+
+	"github.com/cyberark/conjur-authn-k8s-client/pkg/log"
+)
+
+// The function 'envVarValueOrDefault' is defined below. We set this var at the
+// top of the file so when we use the function it is clear that it's the default one.
+var defaultEnvVarValueOrDefaultFunc = envVarValueOrDefault
+
+type EnvVarValueOrDefaultFunc func(envVarName string, defaultValue string) string
```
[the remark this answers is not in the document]
orenbm: made it local

pkg/utils/environment_variable.go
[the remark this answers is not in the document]
sigalsax: So I am understanding correctly if we return an error (for example `return -1, log.RecordedError(log.CAKC010E, envVarName, err.Error())`), no matter what, Golang will zero the return values? So for example, -1 will be 0 because it is an integer?

pkg/utils/environment_variable.go
```diff
@@ -0,0 +1,68 @@
...
+	envVarValueOrDefaultFunc EnvVarValueOrDefaultFunc,
+) (time.Duration, error) {
+	if envVarValueOrDefaultFunc == nil {
+		envVarValueOrDefaultFunc = defaultEnvVarValueOrDefaultFunc
+	}
+
+	envVarValue := envVarValueOrDefault(envVarName, defaultValue)
+	envVarValueDuration, err := time.ParseDuration(envVarValue)
+	if err != nil {
+		return -1, log.RecordedError(log.CAKC010E, envVarName, err.Error())
```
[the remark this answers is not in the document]
sigalsax: which is a bit confusing because usually return 0 = succeeding in bash 🙃

pkg/utils/environment_variable_test.go
sigalsax: @diverdane what are we moving to if not `Convey`

pkg/utils/environment_variable.go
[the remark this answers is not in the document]
orenbm: love it, thanks

pkg/utils/environment_variable.go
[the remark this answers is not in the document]
orenbm: i haven't verified it but in any case we will return 0 to maintain the standard

pkg/utils/environment_variable.go
```diff
@@ -0,0 +1,68 @@
...
+
+type EnvVarValueOrDefaultFunc func(envVarName string, defaultValue string) string
+
+// IntFromEnvOrDefault extracts the value of the given environment variable,
+// or the default value if the environment variable is not defined, and returns
+// it as an int object
+func IntFromEnvOrDefault(
+	envVarName string,
+	defaultValue string,
+	envVarValueOrDefaultFunc EnvVarValueOrDefaultFunc,
```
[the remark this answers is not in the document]
orenbm: i don't see a use-case where an environment variable holds sensitive data so leaving this out. if we will need it in the future we can add another function that will be called from this one for backwards-compatibility. i don't want to over-complicate this util unless i know that it is needed.

pkg/utils/environment_variable_test.go
```diff
@@ -0,0 +1,105 @@
...
+	EnvVarName             = "SOME_ENV_VAR"
+	IntEnvVarValueStr      = "10"
+	IntEnvVarValue         = 10
+	DurationEnvVarValueStr = "10m0s"
+	DurationEnvVarValue    = 10 * time.Minute
+	InvalidValueStr        = "SOME_ENV_VAR_VALUE"
+)
+
+// This mock function always returns the given default value
+func MockEnvVarValueOrDefaultFunc(envVarName string, defaultValue string) string {
```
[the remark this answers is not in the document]
orenbm: thanks for the thorough example!

pkg/authenticator/config/config.go
orenbm: by adding a minimum value we raise a few questions:
1. what is the minimum value of `CONJUR_TOKEN_TIMEOUT`?
2. what is the minimum value of `CONJUR_CLIENT_CERT_RETRY_COUNT_LIMIT`?
3. what do we do in case the given value is less than the minimum? Do we fail the app or use the default value?
4. does the answer to question #3 changes how we handle cases where the user gave an invalid value for `CONJUR_TOKEN_TIMEOUT`?

while these are good questions, i don't want to ask them in this PR and in this release that we are preparing. Once this is merged and the release is out there, we can look back at these questions and decide properly.

orenbm: created an issue for this: https://github.com/cyberark/conjur-authn-k8s-client/issues/162

sigalsax (approved):
LGTM great work!

PR #167, opened 2020-09-15
Add Debug Mode configuration and move detailed logs to Debug log level

### LINKED ISSUE #134: Separate debug logs from info
In cyberark/secrets-provider-for-k8s we have different log levels in which we write messages. This way, log messages that are more for developer debugging can be written only if the user has set the `DEBUG=true` environment variable and the log can be less noisy.

We should add a debug logger and convert any INFO logs that should be logged in debug into DEBUG logs.

DoD:
- [ ] DEBUG log messages are logged only if the `DEBUG=true` env var is presented in the manifest

abrahamko:
### What does this PR do?
- Add configuration to allow Debug mode that will show detailed logs.
  By default, they will not be written, but can be added for investigation.
- Change log level of detailed logs from Info to Debug

Most messages were moved to DEBUG, because INFO should be information on the operation that the invoker is interested in.
For authn client, this is the authentication.
It doesn't matter to him **how** it does it, only **what** it does, and **when** it is done or failed.
See the examples below for more info.

Closes #134 

#### Change log
- [x] The CHANGELOG has been updated

#### Test coverage

#### Documentation

Example of debug mode **off** (default):
```
INFO:  2020/09/15 09:31:55.403986 main.go:16: CAKC014I Kubernetes Authenticator Client v0.18.1-b6d2d1e starting up...
INFO:  2020/09/15 09:31:55.425011 main.go:43: CAKC006I Authenticating as user 'host/conjur/authn-k8s/authn-dev-env/apps/local-secrets-provider/*/*'
INFO:  2020/09/15 09:31:55.665871 main.go:55: CAKC001I Successfully authenticated
INFO:  2020/09/15 09:31:55.665917 main.go:61: CAKC013I Waiting for 6m0s to re-authenticate
```
Example of debug mode **off** (default) *with errors*:
```
INFO:  2020/09/15 09:29:36.001761 main.go:16: CAKC014I Kubernetes Authenticator Client v0.18.1-b6d2d1e starting up...
INFO:  2020/09/15 09:29:36.025154 main.go:43: CAKC006I Authenticating as user 'host/conjur/authn-k8s/authn-dev-env/apps/INVALID/*/*'
ERROR: 2020/09/15 09:29:36.048248 authenticator.go:136: CAKC029E Received invalid response to certificate signing request. Reason: status code 401, 
ERROR: 2020/09/15 09:29:36.048335 authenticator.go:200: CAKC015E Login failed
ERROR: 2020/09/15 09:29:36.048349 main.go:47: CAKC016E Failed to authenticate
```
Example of debug mode **on**
```
INFO:  2020/09/15 09:32:50.828426 main.go:16: CAKC014I Kubernetes Authenticator Client v0.18.1-b6d2d1e starting up...
DEBUG: 2020/09/15 09:32:50.828527 logger.go:60: CAKC001D Debug mode is enabled
INFO:  2020/09/15 09:32:50.894115 main.go:43: CAKC006I Authenticating as user 'host/conjur/authn-k8s/authn-dev-env/apps/local-secrets-provider/*/*'
DEBUG: 2020/09/15 09:32:50.894248 authenticator.go:197: CAKC005I Trying to login Conjur...
DEBUG: 2020/09/15 09:32:50.894299 authenticator.go:116: CAKC007I Logging in as user 'host/conjur/authn-k8s/authn-dev-env/apps/local-secrets-provider/*/*'
DEBUG: 2020/09/15 09:32:50.895002 requests.go:23: CAKC011I Login request to: https://conjur-oss.local-conjur.svc.cluster.local/authn-k8s/authn-dev-env/inject_client_cert
DEBUG: 2020/09/15 09:32:51.074895 authenticator.go:159: CAKC015I Loaded client certificate successfully from /etc/conjur/ssl/client.pem
DEBUG: 2020/09/15 09:32:51.075120 authenticator.go:171: CAKC016I Deleted client certificate from memory
DEBUG: 2020/09/15 09:32:51.075153 authenticator.go:203: CAKC002I Logged in
DEBUG: 2020/09/15 09:32:51.075205 authenticator.go:186: CAKC008I Cert expires: 2020-09-18 09:32:50 +0000 UTC
DEBUG: 2020/09/15 09:32:51.075225 authenticator.go:187: CAKC009I Current date: 2020-09-15 09:32:51.0751843 +0000 UTC
DEBUG: 2020/09/15 09:32:51.075253 authenticator.go:188: CAKC010I Buffer time:  30s
DEBUG: 2020/09/15 09:32:51.075514 requests.go:47: CAKC012I Authn request to: https://conjur-oss.local-conjur.svc.cluster.local/authn-k8s/authn-dev-env/cucumber/host%2Fconjur%2Fauthn-k8s%2Fauthn-dev-env%2Fapps%2Flocal-secrets-provider%2F%2A%2F%2A/authenticate
INFO:  2020/09/15 09:32:51.141853 main.go:55: CAKC001I Successfully authenticated
INFO:  2020/09/15 09:32:51.142157 main.go:61: CAKC013I Waiting for 6m0s to re-authenticate
```


cmd/authenticator/main.go
```diff
@@ -17,6 +17,8 @@ func main() {
 
 	var err error
 
+	configureLogLevel()
```
orenbm, line 20: should this be done in `config.go`?

pkg/authenticator/authenticator.go
```diff
@@ -113,7 +113,7 @@ func (auth *Authenticator) GenerateCSR(commonName string) ([]byte, error) {
 // successfully retrieved
 func (auth *Authenticator) Login() error {
 
-	log.Info(log.CAKC007I, auth.Config.Username)
+	log.Debug(log.CAKC007I, auth.Config.Username)
```
orenbm, line 116: this opens something up. we need to change `CAKC007I` to `CAKC007D` (or some other number) as the last letter indicates the log level. we have a few options here:
1. Change the suffix for every message that we move from Info to Debug
2. Remove the suffice altogether as the level is logged anyway in the same message.
3. Leave these messages with the `I` suffix and not have a `D` suffix at all (info and debug messages will have an `I` suffix).

The second option is the most correct one imo, although it is different from how we log in other projects (conjur, CSPFK). However, option 3 is backwards-compatible and doesn't require a doc change (also - customers may be monitoring these messages).

What do you think?

abrahamko: Good point!
I agree that option 2 looks the most correct way to go, but it made me think: Does it mean that there could be 2 different messages, one for info and one for debug, but with the same ID?
It might create conflicts that will be hard to resolve.
And between 1 and 3, I prefer to go with the convention to decide, which is option 1.

orenbm: they won't have the same ID as one will be `CAKC001I` and one will be `CAKC001D`. we have the same thing with Error log messages. If we go for option 2 then we will need to restructure all the codes from the beginning and it will break compatibility. 

Option 1 is ok as well but it us not fully backwards-compatible. For example, a customer may monitor the log for `CAKC007I` but now it will change to `CAKC007D` and they will never catch it.

Option 3 is the safest one regarding backwards-compatibility. It has its flaws but I would go in that direction.

CHANGELOG.md
```diff
@@ -5,7 +5,13 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Added
+- Configurable debug mode to enable detailed logs.
+  ([cyberark/conjur-authn-k8s-client#134](https://github.com/cyberark/conjur-authn-k8s-client/issues/134))
```
izgeri, line 10:
```suggestion
- Users can set the `DEBUG` environment variable to run the client in debug mode and view more detailed logs.
  [cyberark/conjur-authn-k8s-client#134](https://github.com/cyberark/conjur-authn-k8s-client/issues/134)
```

izgeri (changes requested):
this is great! I look forward to seeing the positive impact on Secretless logs from this change :)

left a couple of requested changes from the changelog

orenbm:
@izgeri the CHANGELOG master!

sgnn7:
@abrahamko / @orenbm / @izgeri Just doing a drive-by comment on this:
While I know that reduction in default logging may be advantageous for smaller logs and less noise, usually the customer provides us with the _default_ log when something goes wrong (which often happens with setups of authn-k8s) and I'm not sure we will get enough detail to troubleshoot those issues quickly and just from the defaults that this PR introduces. I think maybe some contact with support or post-sales would be good to ping on this before we merge the change to see if this is a desired change.

cmd/authenticator/main.go
abrahamko: At first, I thought it shouldn't, as each consumer may enable debug mode however he wants to.
But then I thought it doesn't differ from any other configuration, which done through env vars.
So I will move it to there.

pkg/authenticator/authenticator.go
abrahamko: I just read the [Semantic Versioning](http://semver.org/spec/v2.0.0.html) doc this project adheres to.
It explicitly says: [Major version zero (0.y.z) is for initial development. Anything MAY change at any time. The public API SHOULD NOT be considered stable.](https://semver.org/spec/v2.0.0.html#spec-item-4)
Isn't it enough to allow the use of the 1st option?

abrahamko: BTW, I also found this: https://semver.org/spec/v2.0.0.html#how-do-i-know-when-to-release-100.
@izgeri, why authn client wasn't released as 1.0.0 yet?

orenbm: i think we can go for option 1 but in that case then we can go with option 2. Which one do you prefer?

For both options I would add a CHANGELOG entry saying that the error codes were changed to enhance supportability.

izgeri: @abrahamko I don't know, but I've raised this with PM a couple of times recently. We will have to find a time to have a smallish focused effort on this project to get it to 1.0.0. cc @Tovli

dataplex:
A few comments on this based on some recent conversations:

1. Since we didn't publish a comprehensive list of info/error codes to date for customer consumption, we can change existing ones. We also are pre-version 1.0 so these are subject to change at this time.
2. In the future, we should keep the info/error/debug codes fixed and just create new codes and deprecate the old codes but keep them in the documentation (much like PAS does currently). Handling log message code changes this way makes "breaking" logging code changes easier for customers to adjust to as they may have multiple versions of a component in the environment and migrations may take some time where both sets of log messages need to be handled.
3. I like @orenbm 's comment to "Remove the suffice altogether as the level is logged anyway in the same message."
4. I also share @sgnn7 's concern about not having enough information in the INFO logs currently for troubleshooting purposes. At the very least we should log the configuration INFO when the container is launched to allow us to have those values when troubleshooting or working a ticket. Otherwise we have to reproduce this information by enabling DEBUG, which wastes time on both ends of the interaction (customer/CyberArk support and/or R&D).

izgeri:
Thanks @dataplex! I think we could add a separate issue for adding configuration info logs on container startup - that sounds like a reasonable idea. I would suggest for this PR we:
- Update the log codes to remove the log type suffix (since this is already echoed when the message is logged)
- Review the log messages that will go out once this change is made to ensure there is sufficient info for initial troubleshooting (fwiw this looks reasonable to me, given the process for turning on debug in this container and redeploying is fairly simple - _much_ simpler than in the follower)
- File a follow-up issue to log the initial configuration info on startup

orenbm:
thanks @dataplex & @izgeri !

@abrahamko will you take this from here? when are you planning to do it?

abrahamko:
> thanks @dataplex & @izgeri !
> 
> @abrahamko will you take this from here? when are you planning to do it?
@orenbm I pushed the suffix removal.
@izgeri, you can review the leftover logs in the PR description.

abrahamko:
I opened issue #169 as a follow-up issue to write the configuration to the log.

CHANGELOG.md
```diff
@@ -5,17 +5,23 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Added
+- Users can set the `DEBUG` environment variable to run the client in debug mode and view more log messages.
+  [cyberark/conjur-authn-k8s-client#134](https://github.com/cyberark/conjur-authn-k8s-client/issues/134)
+
 ### Changed
+- Detailed logs moved from Info to Debug log level to decrease verbosity of log messages.
```
izgeri, line 13: we now also need to note here that some error codes changed. maybe we could also link to `pkg/log/log_messages.go`?

pkg/log/log_messages.go
```diff
@@ -13,58 +13,55 @@ package log
...
+const CAKC030 string = "CAKC030 Failed to generate RSA keypair. Reason: %s"
+const CAKC031 string = "CAKC031 Retransmission backoff exhausted"
+const CAKC032 string = "CAKC032 Username %s is invalid"
+const CAKC033 string = "CAKC033 Timed out after waiting for %d seconds for file to exist: %s"
+const CAKC034 string = "CAKC034 Incorrect value '%s' provided for enabling debug mode. Allowed value: '%s'"
+const CAKC035 string = "CAKC035 Successfully authenticated"
+const CAKC036 string = "CAKC036 Logged in"
+const CAKC037 string = "CAKC037 Logged in. Continuing authentication"
+const CAKC038 string = "CAKC038 Certificate expired. Re-logging in..."
+const CAKC039 string = "CAKC039 Trying to login Conjur..."
```
izgeri, line 54:
```suggestion
const CAKC039 string = "CAKC039 Trying to log in to Conjur..."
```

pkg/log/log_messages.go
```diff
@@ -13,58 +13,55 @@ package log
...
-const CAKC012I string = "CAKC012I Authn request to: %s"
-const CAKC013I string = "CAKC013I Waiting for %s to re-authenticate"
-const CAKC014I string = "CAKC014I Kubernetes Authenticator Client v%s starting up..."
-const CAKC015I string = "CAKC015I Loaded client certificate successfully from %s"
-const CAKC016I string = "CAKC016I Deleted client certificate from memory"
-const CAKC017I string = "CAKC017I Waiting for file %s to become available..."
-
-// DEBUG MESSAGES
-const CAKC001D string = "CAKC001D Debug mode is enabled"
+const CAKC001 string = "CAKC001 Error creating the access token object"
```
izgeri, line 16:
```suggestion
const CAKC001 string = "CAKC001 Error opening the access token file"
```

pkg/log/logger.go
```diff
@@ -57,7 +57,7 @@ func Debug(infoMessage string, args ...interface{}) {
 
 func EnableDebugMode() {
```
izgeri, line 58: are there any unit tests we can add for the logger in this effort?

pkg/authenticator/config/config.go
```diff
@@ -62,6 +62,8 @@ func NewFromEnv() (*Config, error) {
 func FromEnv(readFileFunc ReadFileFunc) (*Config, error) {
 	var err error
 
+	configureLogLevel()
```
izgeri, line 65: can we add a unit test for this new method?

izgeri (changes requested):
I left a few comments. It's also worth noting that I ran `golint ./...` against this branch and there were a number of issues - are there any that can be fixed as part of this PR?

orenbm:
@abrahamko can you update on this ticket? where does it stand?
i am adding some new log messages in [another PR](https://github.com/cyberark/conjur-authn-k8s-client/pull/171) and wondered if i should wait for this

abrahamko:
@orenbm I will do one last effort to finish. Otherwise it will have to wait to next R&D boost.

abrahamko:
@izgeri this PR diverged too much from its original purpose.
Please open an issue to address the issues arise from `golint`.

CHANGELOG.md
```diff
@@ -5,17 +5,26 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
...
+  [cyberark/conjur-authn-k8s-client#134](https://github.com/cyberark/conjur-authn-k8s-client/issues/134)
+- Log level suffix was removed from log identifiers (e.g. CAKC001**E** -> CAKC001).
+  See [log_messages.go](https://github.com/cyberark/conjur-authn-k8s-client/blob/master/pkg/log/log_messages.go).
```
izgeri, line 16: it's probably important to note that on account of this, some log identifiers were changed.
```suggestion
- Log level suffix was removed from log identifiers (e.g. `CAKC001**E**` -> `CAKC001`). To
  avoid conflicts, some log identifiers had to be changed. See [log_messages.go](https://github.com/cyberark/conjur-authn-k8s-client/blob/master/pkg/log/log_messages.go)
  for more details.
```

izgeri (dismissed):
I added a suggestion for the change log entry - it's worth calling out that the error messages are changing for some errors with this PR

but other than that, this LGTM, so I'll approve this to remove my "request changes" on the PR

izgeri (approved):

PR #176, opened 2020-10-06
Consume cert injection log only in new Conjur versions

orenbm:
The response code was changed from 200 to 202 in the same Conjur version
that started writing the cert injection logs to the client. Verifying that
the response code is 202 will verify that we look for the log file only
if we expect it to be there.


pkg/authenticator/config/config.go
```diff
@@ -31,7 +31,7 @@ type Config struct {
 // Default settings (this comment added to satisfy linter)
 const (
 	DefaultClientCertPath    = "/etc/conjur/ssl/client.pem"
-	DefaultInjectCertLogPath = "/tmp/conjur_set_file_content.log"
+	DefaultInjectCertLogPath = "/tmp/conjur_copy_text.log"
```
izgeri, line 34: why is this the file name? wouldn't it be clearer to call it something like `cert_injection_error.log` or something?

pkg/authenticator/authenticator.go
```diff
@@ -144,9 +144,15 @@ func (auth *Authenticator) Login() error {
...
 	)
 	if err != nil {
-		injectClientCertError := consumeInjectClientCertError(auth.Config.InjectCertLogPath)
-		if injectClientCertError != "" {
-			log.Error(log.CAKC055, injectClientCertError)
+		// The response code was changed from 200 to 202 in the same Conjur version
+		// that started writing the cert injection logs to the client. Verifying that
+		// the response code is 202 will verify that we look for the log file only
+		// if we expect it to be there
+		if resp.StatusCode == 202 {
```
izgeri, line 151: this is a smart way to do it!

pkg/authenticator/config/config.go
orenbm: the reason is not very strong but it's because the file is created here: https://github.com/cyberark/conjur/blob/master/app/domain/authentication/authn_k8s/copy_text_to_file_in_container.rb#L18

This class is not strictly connected to the cert injection as is quite generic. It gets some text and copies it to a file in the container. I didn't want to call its log file `cert_injection_error.log` so it stays generic (although currently the cert injection is its only use).

what do you think? i'm very open for ideas on this.

pkg/authenticator/authenticator.go
orenbm: Lucky that we also changed the response code :)

pkg/authenticator/config/config.go
izgeri: this particular file is specific to the cert injection, though. it is the `DefaultInjectCertLogPath` - and the file is the cert injection log.

orenbm: yes but the name is set in the server side so the client needs to be aligned with it

orenbm: this is where it is used in the server: https://github.com/cyberark/conjur/blob/master/app/domain/authentication/authn_k8s/copy_text_to_file_in_container.rb#L66

izgeri: ah, gotcha. how about `conjur_server.log`?

orenbm: seems a bit too general

orenbm: @izgeri can we get to a decision on this? i'd like to merge this PR so we can start releasing versions of the different projects.

izgeri: `conjur_authn-k8s_server.log`? I'm not sure where the right balance is between too general (you want this to be a general k8s authn method, right?) and too specific (since this won't _just_ be used for cert injection, theoretically, in the future)

orenbm: yeah, that's why i called it `conjur_copy_text.log`. it indicates that it comes from conjur's action "copy_text".
how about `conjur_server_copy_text_to_file_in_container.log`?
it's quite long but it tells the story.

izgeri: how about `conjur_copy_text_output.log`?

orenbm: sounds good. thanks :)

orenbm: updated the code, please re-review.

i also updated the server side, it will be in another PR

moticless (approved):

PR #178, opened 2020-10-18
Move authentication logs from backoff to auth methods

orenbm:
Some log messages should be part of the authentication methods so that
they are shown also when other projects consume the authn library.

For example, the secrets-provider calls the method `authn.Authenticate()`.
in the current implementation, we will not log the message
`Authenticating as user '%s'` as it's written in the backoff loop which
is not consumed in the secrets provider. Moving these messages into the
methods will let the consuming projects have them.

### DIFF
--- a/pkg/authenticator/authenticator.go
+++ b/pkg/authenticator/authenticator.go
@@ -203,6 +203,8 @@ func (auth *Authenticator) IsCertExpired() bool {
 // Authenticate sends Conjur an authenticate request and returns
 // the response data. Also manages state of certificates.
 func (auth *Authenticator) Authenticate() ([]byte, error) {
+	log.Info(log.CAKC040, auth.Config.Username)
+
 	if !auth.IsLoggedIn() {
 		log.Debug(log.CAKC039)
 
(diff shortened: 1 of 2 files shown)


pkg/authenticator/authenticator.go
```diff
@@ -203,6 +203,8 @@ func (auth *Authenticator) IsCertExpired() bool {
 // Authenticate sends Conjur an authenticate request and returns
 // the response data. Also manages state of certificates.
 func (auth *Authenticator) Authenticate() ([]byte, error) {
+	log.Debug(log.CAKC040, auth.Config.Username)
```
abrahamko, line 206: Maybe add these logs on top of the info logs?
These info logs are the only logs written when debug logs are off, which is the default.

orenbm: i don't understand. would you log these both in INFO and in DEBUG?

InbalZilberman:
@orenbm Can you please link me to the source of these logs (their text :))

orenbm:
> @orenbm Can you please link me to the source of these logs (their text :))

the messages that moved are [this](https://github.com/cyberark/conjur-authn-k8s-client/blob/move-auth-logs/pkg/log/log_messages.go#L50) and [this](https://github.com/cyberark/conjur-authn-k8s-client/blob/move-auth-logs/pkg/log/log_messages.go#L55)

pkg/authenticator/authenticator.go
abrahamko: I think you can add any debug logs you think is necessary, just don't remove the only info logs.

orenbm: ok switched these back to INFO

abrahamko (approved):

PR #179, opened 2020-10-19
Perform all authentication flow in Authenticate

orenbm:
Consumers of the authn-client library currently need to first call the
`Authenticate` method, get the output and call the method
`ParseAuthenticationResponse` with it. This adds redundant complexity
to the flow and can be reduced easily by letting the `Authenticate`
method perform the whole flow by itself. The impact can be seen in
`main.go` (and will affect other consumers such as the secrets-provider
and secretless).

Note that if we perform this change we will need to bump the MAJOR version as this breaks compatibility of the library.

### DIFF
--- a/pkg/authenticator/authenticator.go
+++ b/pkg/authenticator/authenticator.go
@@ -200,16 +200,44 @@ func (auth *Authenticator) IsCertExpired() bool {
 	return currentDate.Add(bufferTime).After(certExpiresOn)
 }
 
-// Authenticate sends Conjur an authenticate request and returns
-// the response data. Also manages state of certificates.
-func (auth *Authenticator) Authenticate() ([]byte, error) {
+// Authenticate sends Conjur an authenticate request and writes the response
+// to the token file (after decrypting it if needed). It also manages state of
+// certificates.
+func (auth *Authenticator) Authenticate() error {
 	log.Info(log.CAKC040, auth.Config.Username)
 
+	err := auth.loginIfNeeded()
+	if err != nil {
+		return err
+	}
+
+	authenticationResponse, err := auth.sendAuthenticationRequest()
+	if err != nil {
+		return err
+	}
+
+	parsedResponse, err := auth.parseAuthenticationResponse(authenticationResponse)
+	if err != nil {
+		return err
+	}
+
+	err = auth.AccessToken.Write(parsedResponse)
+	if err != nil {
+		return err
+	}
+
+	log.Info(log.CAKC035)
+	return nil
+}
+
+// loginIfNeeded checks if we need to send a login request to Conjur and sends
+// one if needed
+func (auth *Authenticator) loginIfNeeded() error {
 	if !auth.IsLoggedIn() {
 		log.Debug(log.CAKC039)
 
 		if err := auth.Login(); err != nil {
-			return nil, log.RecordedError(log.CAKC015)
+			return log.RecordedError(log.CAKC015)
 		}
 
 		log.Debug(log.CAKC036)
@@ -219,12 +247,19 @@ func (auth *Authenticator) Authenticate() ([]byte, error) {
 		log.Debug(log.CAKC038)
 
 		if err := auth.Login(); err != nil {
-			return nil, err
+			return err
 		}
 
 		log.Debug(log.CAKC037)
 	}
 
+	return nil
+}
+
+// sendAuthenticationRequest reads the cert from memory and uses it to send
+// an authentication request to the Conjur server. It also validates the response
+// code before returning its body
+func (auth *Authenticator) sendAuthenticationRequest() ([]byte, error) {
 	privDer := x509.MarshalPKCS1PrivateKey(auth.privateKey)
 	keyPEMBlock := pem.EncodeToMemory(&pem.Block{Type: "RSA PRIVATE KEY", Bytes: privDer})
 
@@ -258,9 +293,9 @@ func (auth *Authenticator) Authenticate() ([]byte, error) {
 	return readBody(resp)
 }
 
-// ParseAuthenticationResponse takes the response from the Authenticate
-// request, decrypts if needed, and writes to the token file
-func (auth *Authenticator) ParseAuthenticationResponse(response []byte) error {
+// parseAuthenticationResponse takes the response from the Authenticate
+// request, decrypts if needed, and returns it
+func (auth *Authenticator) parseAuthenticationResponse(response []byte) ([]byte, error) {
 	var content []byte
 	var err error
 
@@ -268,19 +303,13 @@ func (auth *Authenticator) ParseAuthenticationResponse(response []byte) error {
 	if auth.Config.ConjurVersion == "4" {
 		content, err = decodeFromPEM(response, auth.PublicCert, auth.privateKey)
 		if err != nil {
-			return err
+			return nil, log.RecordedError(log.CAKC020)
 		}
 	} else if auth.Config.ConjurVersion == "5" {
 		content = response
 	}
 
-	err = auth.AccessToken.Write(content)
-	if err != nil {
-		return err
-	}
-
-	log.Info(log.CAKC035)
-	return nil
+	return content, nil
 }
 
 // generateSANURI returns the formatted uri(SPIFFEE format for now) for the certificate.
--- a/pkg/authenticator/authenticator.go
+++ b/pkg/authenticator/authenticator.go
@@ -131,7 +131,7 @@ func (auth *Authenticator) Login() error {
 		return log.RecordedError(log.CAKC028, err)
 	}
 
-	err = EmptyResponse(resp)
+	err = validateResponse(resp)
 	if err != nil {
 		return log.RecordedError(log.CAKC029, err)
 	}
@@ -250,7 +250,12 @@ func (auth *Authenticator) Authenticate() ([]byte, error) {
 		return nil, log.RecordedError(log.CAKC027, err)
 	}
 
-	return DataResponse(resp)
+	err = validateResponse(resp)
+	if err != nil {
+		return nil, err
+	}
+
+	return readBody(resp)
 }
 
 // ParseAuthenticationResponse takes the response from the Authenticate
--- a/pkg/authenticator/requests.go
+++ b/pkg/authenticator/requests.go
@@ -55,6 +55,7 @@ func AuthenticateRequest(authnURL string, conjurVersion string, account string, 
 	return req, nil
 }
 
+// readBody returns the response body
 func readBody(resp *http.Response) ([]byte, error) {
 	defer resp.Body.Close()
 
@@ -66,20 +67,10 @@ func readBody(resp *http.Response) ([]byte, error) {
 	return responseBytes, err
 }
 
-// DataResponse checks the HTTP status of the response. If it's less than
+// validateResponse checks the HTTP status of the response. If it's less than
 // 300, it returns the response body as a byte array. Otherwise it returns
 // a NewError.
-func DataResponse(resp *http.Response) ([]byte, error) {
-	if resp.StatusCode < 300 {
-		return readBody(resp)
-	}
-	return nil, NewError(resp)
-}
-
-// EmptyResponse checks the HTTP status of the response. If it's less than
-// 300, it returns without an error. Otherwise it returns
-// a NewError.
-func EmptyResponse(resp *http.Response) error {
+func validateResponse(resp *http.Response) error {
 	if resp.StatusCode < 300 {
 		return nil
 	}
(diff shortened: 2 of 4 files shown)


abrahamko:
I'm all in for combining the two functions, but this breaks the API with no backwards compatibility.
Besides hiding the `ParseAuthenticationResponse`, you changed the returned value of `Authenticate`.
If we had erleased authn client, we had to increase MAJOR VERSION number for such change.
But we're not, so it ok, I guess.
Simpler approach would be to introduce a new function that wraps it, so you don't break, and I would recommend it if it was released.
@izgeri WDYT?

abrahamko (changes requested):

pkg/authenticator/requests.go
```diff
@@ -66,10 +66,10 @@ func readBody(resp *http.Response) ([]byte, error) {
 	return responseBytes, err
 }
 
-// DataResponse checks the HTTP status of the response. If it's less than
+// ValidatedResponse checks the HTTP status of the response. If it's less than
 // 300, it returns the response body as a byte array. Otherwise it returns
 // a NewError.
-func DataResponse(resp *http.Response) ([]byte, error) {
+func ValidatedResponse(resp *http.Response) ([]byte, error) {
```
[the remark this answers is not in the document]
orenbm: the motivation was to give it a better name. Not sure why this method needs to be public anyway.
i'm ok with breaking the API as the rest of the PR does. 

let's wait for others to comment on the general change

orenbm:
Converted this to a draft as we need to change the changelog entry

pkg/authenticator/authenticator.go
```diff
@@ -242,40 +243,50 @@ func (auth *Authenticator) Authenticate() ([]byte, error) {
...
 
 	resp, err := client.Do(req)
 	if err != nil {
-		return nil, log.RecordedError(log.CAKC027, err)
+		return log.RecordedError(log.CAKC027, err)
+	}
+
+	validatedResponse, err := ValidatedResponse(resp)
+	if err != nil {
+		return err
```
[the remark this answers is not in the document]
orenbm: fixed thanks

pkg/authenticator/requests.go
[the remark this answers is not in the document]
abrahamko: The validation part is a side effect of this method.
The signature should explain what is the meaning of the returned value (byte array).
So maybe `getDataFromResponse` is a better name?
Also, make it private if it is not needed to be public.

orenbm:
@cyberark/community-and-integrations-team can we assume that the only consumers of this library are secretless-broker and secrets-provider?

pkg/authenticator/requests.go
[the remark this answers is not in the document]
doodlesbykumbi: Here are my thoughts:

1. I'm not sure if this method needs to be public
2. I see that both `ValidatedResponse` and `EmptyResponse` are sort of doing the same thing but their names are focusing on the wrong thing. I think really we want 2 functions, 1 that validates the responses, and 1 that reads the body from the response.

I'd say:
1. Delete `EmptyResponse` and `ValidatedResponse`
2. Add `validateResponse`
```
func validateResponse(resp *http.Response) error {
	if resp.StatusCode < 300 {
		return nil
	}
	return NewError(resp)
}
```
3. Where `EmptyResponse` was used simply use `validateResponse`
4. Where `ValidatedResponse` was used simply `validateResponse`. followed by `readBody`

pkg/authenticator/authenticator.go
```diff
@@ -242,40 +243,50 @@ func (auth *Authenticator) Authenticate() ([]byte, error) {
...
 	}
 
-	return DataResponse(resp)
+	parsedResponse, err := auth.parseAuthenticationResponse(validatedResponse)
```
doodlesbykumbi, line 259: I wonder if it would help with readability to move everything in this function (Authenticate) above this line to a function `makeAuthenticationRequest`. 

This would then leave `Authenticate` looking something like this, where the `...` is logging and error handling:
```
func (auth *Authenticator) Authenticate() error {
    ...
    auth.makeAuthenticationRequest
    ...
    auth.parseAuthenticationResponse
    ...
    auth.AccessToken.Write
    ...
}
```

doodlesbykumbi (commented):
Left some comments. I think the motivation behind the change is valid. The public API really should just allow a single call to retrieve the decrypted token.

orenbm:
@doodlesbykumbi @abrahamko can you please re-review?

doodlesbykumbi (approved):
LGTM! Approved for my concerns, I'll leave it for the rest.

pkg/authenticator/authenticator.go
```diff
@@ -200,16 +200,44 @@ func (auth *Authenticator) IsCertExpired() bool {
...
+
+// loginIfNeeded checks if we need to send a login request to Conjur and sends
+// one if needed
+func (auth *Authenticator) loginIfNeeded() error {
 	if !auth.IsLoggedIn() {
 		log.Debug(log.CAKC039)
 
 		if err := auth.Login(); err != nil {
-			return nil, log.RecordedError(log.CAKC015)
+			return log.RecordedError(log.CAKC015)
```
sgnn7, line 240: Not a blocker but it's a bit strange that we don't use the returned errors for neither `CAKC015` nor `CAKC020` in any way

sgnn7 (changes requested):
@orenbm Couple of notes:
- Only one comment about the code itself. There's some throwing away of error data that we probably want to log.
- This change seems like it should have at least some tests added.
- This is definitely a breaking change in APIs. I will leave @izgeri to comment on versioning with it but I feel like we may need to preserve backward compatibility until next major version.

pkg/authenticator/authenticator.go
```diff
@@ -250,32 +285,31 @@ func (auth *Authenticator) Authenticate() ([]byte, error) {
...
-func (auth *Authenticator) ParseAuthenticationResponse(response []byte) error {
+// parseAuthenticationResponse takes the response from the Authenticate
+// request, decrypts if needed, and returns it
+func (auth *Authenticator) parseAuthenticationResponse(response []byte) ([]byte, error) {
 	var content []byte
 	var err error
 
 	// Token is only encrypted in Conjur v4
 	if auth.Config.ConjurVersion == "4" {
 		content, err = decodeFromPEM(response, auth.PublicCert, auth.privateKey)
```
sigalsax, line 304: why are we decoding only in v4 and not also in v5?

orenbm: i guess v4 returned the certificate encoded and v5 does not. i hope this code snippet can be removed soon

pkg/authenticator/authenticator.go
```diff
@@ -219,12 +247,19 @@ func (auth *Authenticator) Authenticate() ([]byte, error) {
...
 		log.Debug(log.CAKC037)
 	}
 
+	return nil
+}
+
+// makeAuthenticationRequest reads the cert from memory and uses it to send
+// an authentication request to the Conjur server. It also validates the response
+// code before returning its body
+func (auth *Authenticator) makeAuthenticationRequest() ([]byte, error) {
```
abrahamko, line 262: Good method separation!
IMO `sendAuthenticationRequest` is better than `makeAuthenticationRequest`
```suggestion
func (auth *Authenticator) sendAuthenticationRequest() ([]byte, error) {
```

pkg/authenticator/authenticator.go
```diff
@@ -250,32 +285,31 @@ func (auth *Authenticator) Authenticate() ([]byte, error) {
 		return nil, log.RecordedError(log.CAKC027, err)
 	}
 
-	return DataResponse(resp)
+	err = validateResponse(resp)
+	if err != nil {
+		return nil, err
+	}
+
+	return readBody(resp)
```
abrahamko, line 293: `readBody` is a private function in another file.
Just a thought...
I know it's valid `go` because they reside in the same package (which means the same folder in out convention), but is less readable IMO.
Maybe it should be moved to a different package? 🤔

pkg/authenticator/authenticator.go
```diff
@@ -200,16 +200,44 @@ func (auth *Authenticator) IsCertExpired() bool {
...
+	}
+
+	authenticationResponse, err := auth.makeAuthenticationRequest()
+	if err != nil {
+		return err
+	}
+
+	parsedResponse, err := auth.parseAuthenticationResponse(authenticationResponse)
+	if err != nil {
+		return log.RecordedError(log.CAKC020)
```
abrahamko, line 221: All the other methods called in this function does not log the error returned.
Maybe this log should be moved into the `parseAuthenticationResponse` implementation?

abrahamko (commented):
Left some comments.
LGTM!

pkg/authenticator/authenticator.go
izgeri: we deprecated support for authn-k8s in [DAP 11.5](https://docs.cyberark.com/Product-Doc/OnlineHelp/PAS/Latest/en/Content/Release%20Notes/RN-WhatsNew11-5.htm?tocpath=Get%20Started%7CWhat%E2%80%99s%20New%7CRelease%20Notes%7C_____5)

I think we can remove support in the next release, since we already announced deprecation with 11.5. the changelog entry should probably reference the 11.5 release notes. you can see the [Puppet changelog](https://github.com/cyberark/conjur-puppet/blob/master/CHANGELOG.md) for example entries on removing support.

izgeri:
@sgnn7 how big is the breaking change in the API? this project is still pre-1.0, so I think that's ok - I just want to understand the impact on the downstream projects. the changelogs should also be super clear about this.

sgnn7:
@izgeri The usage would just absorb some code that was needed on the client side into the lib itself. In general it's a tiny API change from the consumer's point of view afaict.

orenbm:
@izgeri @sgnn7 are we good to go with this change?

pkg/authenticator/authenticator.go
orenbm: nice, thanks

pkg/authenticator/authenticator.go
orenbm: wdym? we log the message here and also raise it in the `Authenticate()` method

pkg/authenticator/authenticator.go
orenbm: i think it makes sense to have it in the `requests` package

pkg/authenticator/authenticator.go
orenbm: Created an issue for this - https://github.com/cyberark/conjur-authn-k8s-client/issues/183

izgeri:
@orenbm I am not sure we should move to 1.0.0 - it would be nice to reserve that for a time when we can do a push to review / revise / refactor the full project using the standard definition of done we have. I am concerned we may release 1.0 and then find there are other breaking changes we would want to make - do you have capacity to revisit this project, review all changes to make before 1.0, and implement those changes soon?

orenbm:
> @orenbm I am not sure we should move to 1.0.0 - it would be nice to reserve that for a time when we can do a push to review / revise / refactor the full project using the standard definition of done we have. I am concerned we may release 1.0 and then find there are other breaking changes we would want to make - do you have capacity to revisit this project, review all changes to make before 1.0, and implement those changes soon?

Yeah i actually thought about it later and it seems we can stay at major version zero even if we break the API. this is from SemVer:
```
Major version zero is all about rapid development. If you’re changing the API every day you should either still be in version 0.y.z or on a separate development branch working on the next major version.
```.

So i would not bump to 1.0.0 with this change.

orenbm:
@izgeri @abrahamko @sgnn7 can we merge this PR?

pkg/authenticator/authenticator.go
abrahamko: Do you plan to move it to there?

abrahamko (commented):
LGTM.
I will let @sgnn7 / @izgeri approve the API breaking.

pkg/authenticator/authenticator.go
abrahamko: Do you plan to move it to there?

orenbm: it's already there :)

izgeri:
it's fine for the API to break, since we are pre-1.0. please make sure it's clear that it's breaking in the changelog message, and that the linked issue has enough info about _how_ it changed that downstream project maintainers can figure out how to update their apps. we should also internally alert teams that use this library that the next update will require some API changes.

orenbm:
> * This change seems like it should have at least some tests added.

@sgnn7 i am not adding UTs in this PR for the following reasons:
- `Authenticate`: this method is quite big and a UT can be complex. I created [an issue](https://github.com/cyberark/conjur-authn-k8s-client/issues/185) for this
- `parseAuthenticationResponse`: this method will be deleted very soon as its logic is mainly for Conjur v4 which will be removed

abrahamko (approved):
Since @izgeri approved the API break, this LGTM.

pkg/authenticator/authenticator.go
orenbm: created a PR for this - https://github.com/cyberark/conjur-authn-k8s-client/pull/187

we need to add UTs so it's not ready yet and i have other pressing matters. you can review the changes in your spare time, thanks!

PR #184, opened 2020-10-28
Rename injection to placement in log message

orenbm:
The term "injection" has a negative connotation. We would like to use a term with a more positive connotation.


InbalZilberman (approved):
Thank you!

sigalsax (approved):

PR #195, opened 2020-11-10
Remove temporary TAG value

orenbm:
There is no use in setting the TAG argument to `dev` as it is set in
`./bin/build`. It is more accurate to leave it blank and say that it is
set in `./bin/build` rather than setting it to some value and saying
that the value will change.

This change will reduce understandment complexity.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,8 +5,8 @@ ENV GOOS=linux \
     GOARCH=amd64 \
     CGO_ENABLED=1
 
-# this value changes in ./bin/build
-ARG TAG="dev"
+# this value is set in ./bin/build
+ARG TAG
 
 WORKDIR /opt/conjur-authn-k8s-client
 COPY . /opt/conjur-authn-k8s-client


eladkug (approved):

sgnn7:
@orenbm The reason why that default value is there is so that there is a default value when building the docker container without that script which results in version `v1.2.3-` when built which is somewhat of a bug (if I'm mis-remembering here, feel free to correct me). If the tag value is correctly applied by `build` script, then there should be no effective difference.

orenbm:
> @orenbm The reason why that default value is there is so that there is a default value when building the docker container without that script 

@sgnn7 can you please elaborate on this? which command do you run to do this? what is the use-case in which you will run this and not run the `./bin/build` script?

sgnn7:
@orenbm It's a community project so anyone that sees `Dockerfile` and does `docker build .` will encounter this odd `v1.2.3-` version in their image with this change. It's not a blocker but it does make for a slightly worse UX for a new developer that doesn't know that `./bin/build` needs to be used to build the image. Since the PR has an approval, it's up to you what you want to do, I commented to just add the reason why the original default value was there for.

orenbm:
nah, the value in this change is not big enough. closing.

PR #209, opened 2021-01-20
Fix duplicated word in log message

orenbm:


abrahamko (approved):
I checked that secrets provider tests does not depend on this string.
LGTM!

PR #212, opened 2021-02-01
Add reason to invalid username error message

orenbm:
### What does this PR do?
We check that the username starts with 'host/' so we should write this
in the error message for better usability.


pkg/log/log_messages.go
```diff
@@ -44,7 +44,7 @@ const CAKC028 string = "CAKC028 Failed to send https login request or response.
 const CAKC029 string = "CAKC029 Received invalid response to certificate signing request. Reason: %s"
 const CAKC030 string = "CAKC030 Failed to generate RSA keypair. Reason: %s"
 const CAKC031 string = "CAKC031 Retransmission backoff exhausted"
-const CAKC032 string = "CAKC032 Username %s is invalid"
+const CAKC032 string = "CAKC032 Username %s is invalid. Verify it start with 'host/'."
```
orenbm, line 47: @shulifink please review

pkg/log/log_messages.go
```diff
@@ -44,7 +44,7 @@ const CAKC028 string = "CAKC028 Failed to send https login request or response.
 const CAKC029 string = "CAKC029 Received invalid response to certificate signing request. Reason: %s"
 const CAKC030 string = "CAKC030 Failed to generate RSA keypair. Reason: %s"
 const CAKC031 string = "CAKC031 Retransmission backoff exhausted"
-const CAKC032 string = "CAKC032 Username %s is invalid"
+const CAKC032 string = "CAKC032 Username %s is invalid. Verify it starts with 'host/'"
```
mbenita-Cyberark, line 47: Could be a bit misleading. in case the username starts with a host and the username is still invalid. the user can get confused. 
I'm not farmilliar with the code style, but i see two options:
1) check if it doesn't starts with host/ and raise this message
2) `const CAKC032 string = "CAKC032 Username %s is invalid. Common Error - Username doesn't starts with 'host/'"`

orenbm: We actually check only that the username starts with `host/` as you can see here: https://github.com/cyberark/conjur-authn-k8s-client/blob/master/pkg/authenticator/config/username.go#L37.

I will add a method in the calling code to make it clearer for future developers

orenbm: how about changing the text to `CAKC032 Username %s must start with 'host/'`

abrahamko: The name `username` is misleading by itself.
It is originated from `CONJUR_AUTHN_LOGIN` - what's the point of talking about username?
better talk about `login` or `login_id` or anything that explains the field this issue refers to.

abrahamko: Also, why failing?
We can get better UX by adding `host` ourselves, no?
As a guideline, I think we shouldn't fail if we don't have to; be more graceful.

mbenita-Cyberark: I don't farmilliar enough with go, or the authenticator to address @abrahamko comment

> how about changing the text to `CAKC032 Username %s must start with 'host/'`?

Looks like the best suited error message to this case.

izgeri: I think @abrahamko's feedback is good. I'd suggest `CAKC032 CONJUR_AUTHN_LOGIN %s must start with 'host/'` if the value always comes from `CONJUR_AUTHN_LOGIN`, otherwise maybe something simple like `Conjur login ID`?

orenbm: thanks everyone, fixed according to your feedback.

abrahamko: @orenbm did you consider my last comment?
> Also, why failing?
We can get better UX by adding host ourselves, no?
As a guideline, I think we shouldn't fail if we don't have to; be more graceful.

orenbm: i'm quite afraid to make such a change without fully looking into it. what if the customer tries to use a user? in this case we will add `host/` to it and it may fail because of some other unxpected behaviour.

abrahamko: I think this is something we should consider in the future, not relevant for now.

abrahamko (approved):

pkg/log/log_messages.go
orenbm: thanks, created an issue: https://github.com/cyberark/conjur-authn-k8s-client/issues/214

PR #254, opened 2021-03-23
Adding the cluster prep Helm chart

### LINKED ISSUE #232: There is an initial cluster prep helm chart
## Overview
This user story requires the definition of a cluster prep helm chart, including:
- `values.yaml` file
- `templates/golden_configmap.yaml` templated manifest
- `templates/service_account.yaml` templated manifest
- `templates/clusterrole.yaml` templated manifest
- `README.md` explaining basic usage and a table of configurable values
- templates/`NOTES.txt` simply announcing completion, maybe listing objects that have been deployed

### Out of Scope:
- `values.schema.json` (covered in Issue #233)
- Unit test based on `unittest` framework (covered in Issue #234)
- Helm test validation (covered in Issue #228)

### Kubernetes Objects to be Created for Helm install/upgrade
When `helm install ...` is invoked on this Helm chart, there should be two Kubernetes objects created in the authenticator Namespace:
- Golden (Reference) ConfigMap
- Authenticator ServiceAccount
- Authenticator ClusterRole

### Sample Golden ConfigMap Manifest
When `helm install ...` is invoked using this Helm chart, Helm should render a manifest for a Golden ConfigMap that looks similar to the following:

```
apiVersion: v1
kind: ConfigMap
metadata:
  name: authn-k8s-config-map
  labels:
    app.kubernetes.io/name: golden-configmap
    app.kubernetes.io/component: conjur-config
    app.kubernetes.io/instance: <Authenticator Namespace>
    app.kubernetes.io/part-of: authn-k8s-cluster-config
    app.kubernetes.io/managed-by: helm
    helm.sh/chart: <Chart name>-<Chart version>
data:
  # authn-k8s Configuration
  authnK8sAuthenticatorID: my-authenticator-id
  authnK8sClusterRole: conjur-authenticator-clusterrole
  authnK8sNamespace: conjur-ns
  authnK8sServiceAccount: conjur-sa

  # Conjur Configuration
  conjurAccount: myConjurAccount
  conjurApplianceUrl: https://conjur.conjur-ns.svc.cluster.local
  conjurSslCertificate: <Unencoded Conjur SSL certificate>
  conjurSslCertificateBase64: <base64 encoded Conjur SSL certificate>
```

### Sample ClusterRole Manifest
When `helm install ...` is invoked using this Helm chart, Helm should render a manifest for a ClusterRole that looks similar to the following:
```
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  labels:
    app.kubernetes.io/name: authn-k8s-clusterrole
    app.kubernetes.io/component: authn-k8s-permissions
    app.kubernetes.io/instance: <Authenticator Namespace>
    app.kubernetes.io/part-of: authn-k8s-cluster-config
    app.kubernetes.io/managed-by: helm
    helm.sh/chart: <Chart name>-<Chart version>
  name: authn-k8s
rules:
- apiGroups:
  - ""
  resources:
  - pods
  - serviceaccounts
  verbs:
  - get
  - list
- apiGroups:
  - extensions
  resources:
  - deployments
  - replicasets
  verbs:
  - get
  - list
- apiGroups:
  - apps
  resources:
  - deployments
  - statefulsets
  - replicasets
  verbs:
  - get
  - list
- apiGroups:
  - ""
  resources:
  - pods/exec
  verbs:
  - create
  - get
```

### Sample ServiceAccount Manifest
When `helm install ...` is invoked using this Helm chart, Helm should render a manifest for a ServiceAccount that looks similar to the following:
```
apiVersion: v1
kind: ServiceAccount
metadata:
  labels:
    app.kubernetes.io/name: authn-k8s-serviceaccount
    app.kubernetes.io/component: authn-k8s-identity
    app.kubernetes.io/instance: <Authenticator Namespace>
    app.kubernetes.io/part-of: authn-k8s-cluster-config
    app.kubernetes.io/managed-by: helm
    helm.sh/chart: <Chart name>-<Chart version>
  name: authn-k8s
```

### Helm chart Files to be created
This Helm chart requires the creation of the following files in the chart subdirectory:
- `values.yaml` file
- `templates/golden_configmap.yaml` templated manifest
- `templates/authn_serviceaccount.yaml` templated manifest
- `templates/authn_clusterrole.yaml` templated manifest
- `README.md` explaining basic usage, upgrade examples, and a table of configurable values
- templates/`NOTES.txt` simply announcing completion, maybe listing objects that have been deployed

### `values.yaml` requirements
The `values.yaml` file created for this Helm chart must contain the fields that are listed in
the "Kubernetes Cluster Prep Helm Chart Value" column of the
"User Input: Information Gathered for Kubernetes Cluster Preparation" section of this
document: https://cyberark365.sharepoint.com/:w:/s/Conjur/EbhBd0DKYt1MstzSj-4sWOkBssiD23FAHA4JgW-2V2YwrA?e=04gdla.

### Additional "control knobs"
The `values.yaml` file should also include a couple of other "control knobs":
- authnK8s.rbac.enable
   This is needed to enable/disable creation of the authn-k8s ClusterRole.
   This is future-proofing: There may be cases in the future whereby someone might want
   to share ClusterRoles across multiple authn-k8s authenticators.
- authnK8s.serviceAccount.enable
   This is for future-proofing. For some deployments, e.g. with the Conjur OSS Helm chart,
   there already exists a ServiceAccount that can be used/re-used.
 
 ### Default values
The `values.yaml` file should provide the following default values:
- `conjur.account` defaults to "default"
- `authnK8s.rbac.enable` defaults to true
- `authnK8s.serviceAccount.enable` defaults to true
 
### Template manifests requirements
The template manifests must enforce that required values have been explicitly defined.
Required values are those fields in the `values.yaml` file that are indicated as
"Mandatory" in the "Kubernetes Cluster Prep Helm Chart Value" column of the
"User Input: Information Gathered for Kubernetes Cluster Preparation" section of this
document: https://cyberark365.sharepoint.com/:w:/s/Conjur/EbhBd0DKYt1MstzSj-4sWOkBssiD23FAHA4JgW-2V2YwrA?e=04gdla.

### `README.md` requirements 
The `README.md` must include:
- Overview of the intended use of the chart.
- Step-by-step instructions on how to use the Helm chart, including:
   - Prequisite information that needs to be gathered from a Conjur admin (not including Conjur SSL cert).
   - Instructions on how to retrieve the Conjur SSL certificate based on the Conjur URL
   - Table of configurable values
   - Examples on how Helm upgrade can be used to modify values used by a chart

### `NOTES.txt` requirements
`NOTES.txt` should announce completion, maybe listing objects that have been deployed

## Environments on Which to Test

For this implementation, it should be sufficient to test this Helm chart on:
- Conjur OSS
- Either Kubernetes cluster (can be KinD or GKE) or OpenShift, pick one

DoD:
- Required files have been created
- Helm chart tested on setups described in "Environments on Which to Test" above
- Helm install creates Golden ConfigMap, ServiceAccount, ClusterRole using default enable/disable flags
- Disabling of ServiceAccount creation works for helm install
- Disabling of ClusterRole creation works for helm install
- Helm upgrade works as expected
- Requirements listed above are all met

rpothier:
### What does this PR do?

Initial adding of the cluster prep Helm chart

Resolves #232 

#### Change log

#### Test coverage

#### Documentation

#### Manual tests


deploy/charts/kubernetes-cluster-prep/.helmignore
```diff
@@ -0,0 +1,27 @@
...
+*.bak
+*.tmp
+*.orig
+*~
+# Various IDEs
+.project
+.idea/
+*.tmproj
+.vscode/
+# Helm test image build
```
diverdane, line 24: We can probably leave this out for now... this is for the 'helm test' W.I.P. / P.o.C. that I'm working on.

deploy/charts/kubernetes-cluster-prep/Chart.yaml
```diff
@@ -0,0 +1,21 @@
+apiVersion: v1
+name: conjur-cluster-prep
+description: A Helm chart for Kubernetes
+
```
diverdane, line 4: Please add:
```
icon: https://www.cyberark.com/wp-content/uploads/2015/12/cybr-aim.jpg
```
after description line.

deploy/charts/kubernetes-cluster-prep/Chart.yaml
```diff
@@ -0,0 +1,21 @@
+apiVersion: v1
+name: conjur-cluster-prep
+description: A Helm chart for Kubernetes
```
diverdane, line 3: Let's make description:
```
description: A Helm chart for preparing a Kubernetes cluster to use
             the Conjur Kubernetes authenticator (authn-k8s)
```

deploy/charts/kubernetes-cluster-prep/Chart.yaml
```diff
@@ -0,0 +1,21 @@
+apiVersion: v1
```
diverdane, line 1: Let's make this `v2`, I think we need that for unittest.

deploy/charts/kubernetes-cluster-prep/README.md
```diff
@@ -0,0 +1,61 @@
+# Helm chart for Conjur Kubernetes cluster prep
```
diverdane, line 1: NICE README.md!!!!!  I'll look at this in more depth after reviewing the other files.

diverdane: We'll need a table with all configurable settings.
There's an example in cyberark/conjur-oss-helm-chart/conjur-oss/README.txt.

deploy/charts/kubernetes-cluster-prep/sample-policy/cluster-authn-svc-def.yaml
```diff
@@ -0,0 +1,36 @@
+---
```
diverdane, line 1: You can delete the entire `sample-policy` directory. This will go in with the 'helm test'.

deploy/charts/kubernetes-cluster-prep/templates/NOTES.txt
```diff
@@ -0,0 +1,31 @@
...
+A Golden ConfigMap
+{{ end }}
+{{ if .Values.authnK8s.rbac.create }}
+An authenticator ClusterRole
+{{ end }}
+{{ if .Values.authnK8s.serviceAccount.create }}
+An authenticator ServiceAccount
+{{ end }}
+
+{{ if .Values.displaySamplePolicy }}
```
diverdane, line 13: Please delete everything from 13 to the end of file. This will get added with the `helm test` commit.

diverdane (commented):
Looking good!!!
I still have more reviewing to do, but so-far, so-good.

deploy/charts/kubernetes-cluster-prep/templates/clusterrole.yaml
```diff
@@ -0,0 +1,28 @@
+{{- if .Values.authnK8s.rbac.create }}
```
diverdane, line 1: I think it makes sense to get rid of the `rbac` and use these settings in `values.yaml`, so that we're consistent with the other settings in `values.yaml`:
```
authnK8s:
  clusterRole:
    create: true
    name:
```
If you agree, then this line would be:
```
{{- if .Values.authnK8s.clusterRole.create }}
```

deploy/charts/kubernetes-cluster-prep/.helmignore
rpothier: oh right, I went right past this. Will remove.

deploy/charts/kubernetes-cluster-prep/templates/clusterrole.yaml
```diff
@@ -0,0 +1,28 @@
...
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: {{ .Values.authnK8s.rbac.clusterRole }}
```
diverdane, line 5: If we get rid of the `rbac`, this line would be:
```
  name: {{ .Values.authnK8s.clusterRole.name }}
```

deploy/charts/kubernetes-cluster-prep/templates/clusterrole.yaml
```diff
@@ -0,0 +1,28 @@
...
+  labels:
+    release: {{ .Release.Name }}
+    heritage: {{ .Release.Service }}
+    app.kubernetes.io/name: {{ .Release.Name }}
```
diverdane, line 9: I *think* this should be a more generic, fixed name, e.g.:
```
    app.kubernetes.io/name: authn-k8s-clusterrole
```
based on this reference: https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/#labels

Apologies, I think the issue and the design spec were not quite consistent with Kubernetes guidelines.

deploy/charts/kubernetes-cluster-prep/Chart.yaml
rpothier: sure, just curious, where does this appear?

deploy/charts/kubernetes-cluster-prep/templates/clusterrole.yaml
```diff
@@ -0,0 +1,28 @@
...
+    app.kubernetes.io/component: conjur-config
```
diverdane, line 10: Looking at the Kubernetes doc noted earlier, the `component` is a high level description of where this thing fits in the architecture. So maybe:
```
    app.kubernetes.io/component: authn-k8s-permissions
```
or:
```
    app.kubernetes.io/component: rbac
```

deploy/charts/kubernetes-cluster-prep/templates/clusterrole.yaml
```diff
@@ -0,0 +1,28 @@
...
+    app.kubernetes.io/instance: {{ .Release.Namespace }}
+    app.kubernetes.io/part-of: "authn-k8s-cluster-config"
```
diverdane, line 12: Let's add this:
```
    app.kubernetes.io/managed-by: helm
```
since that's another standard label, might be useful to someone/something.

deploy/charts/kubernetes-cluster-prep/templates/golden_configmap.yaml
```diff
@@ -0,0 +1,41 @@
+{{- if .Values.authnK8s.configMap.create -}}
+apiVersion: v1
+kind: ConfigMap
+metadata:
+    name: {{ .Values.authnK8s.configMap.name }}
+    labels:
+      release: {{ .Release.Name }}
+      heritage: {{ .Release.Service }}
+      app.kubernetes.io/name: {{ .Values.authnK8s.configMap.name }}
```
diverdane, line 9: Should be something fixed so we can search on it, e.g.:
```
      app.kubernetes.io/name: golden-configmap
```

deploy/charts/kubernetes-cluster-prep/templates/golden_configmap.yaml
```diff
@@ -0,0 +1,41 @@
...
+      app.kubernetes.io/component: "conjur-config"
+      app.kubernetes.io/instance: {{ .Release.Namespace }}
+      app.kubernetes.io/part-of: "authn-k8s-cluster-config"
```
diverdane, line 12: Let's add this:
```
    app.kubernetes.io/managed-by: helm
```

diverdane (commented):
More comments.

deploy/charts/kubernetes-cluster-prep/templates/NOTES.txt
```diff
@@ -0,0 +1,31 @@
+The Conjur/Authenticator Namespace preparation is complete.
+The following have been deployed:
+{{ if .Values.authnK8s.configMap.create }}
+A Golden ConfigMap
+{{ end }}
+{{ if .Values.authnK8s.rbac.create }}
```
diverdane, line 6: See note below, this should be changed to:
```
{{ if .Values.authnK8s.clusterRole.create }}
```

deploy/charts/kubernetes-cluster-prep/templates/golden_configmap.yaml
```diff
@@ -0,0 +1,41 @@
...
+      helm.sh/chart: {{ .Chart.Name }}-{{ .Chart.Version }}
+data:
+    # authn-k8s Configuration
```
diverdane, line 15: I've been making changes to get unittest to work. Probably easiest to lift the rest of this configmap data section from here:
https://github.com/diverdane/conjur-cluster-prep/blob/master/templates/golden_configmap.yaml#L7

deploy/charts/kubernetes-cluster-prep/templates/clusterrole.yaml
rpothier: yes, I like that better.

deploy/charts/kubernetes-cluster-prep/templates/serviceaccount.yaml
```diff
@@ -0,0 +1,14 @@
+{{- if .Values.authnK8s.serviceAccount.create -}}
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: {{ .Values.authnK8s.serviceAccount.name }}
+  labels:
+    release: {{ .Release.Name }}
+    heritage: {{ .Release.Service }}
+    app.kubernetes.io/name: {{ .Release.Name }}
```
diverdane, line 9: Maybe make these more generic/fixed:
```
    app.kubernetes.io/name: authn-k8s-serviceaccount
    app.kubernetes.io/component: authn-k8s-identity
```
or just `identity` for the last one?

deploy/charts/kubernetes-cluster-prep/values.yaml
```diff
@@ -0,0 +1,53 @@
+# Default values for conjur-cluster-prep.
```
diverdane, line 1: I made a bunch of changes to get the upcoming changes for unittest and schema validation to work.
Maybe easiest to copy this file:
https://github.com/diverdane/conjur-cluster-prep/blob/master/values.yaml

diverdane: Mind adding this script?:
https://github.com/diverdane/conjur-cluster-prep/blob/master/test-lint
(This might help you check for errors as you make changes.)

diverdane (commented):
I've reviewed everything but the `README.md`. Looking at that now.

deploy/charts/kubernetes-cluster-prep/values.yaml
diverdane: If you do add the `test-lint` script, you'll also have to add the `tests/test-cert.pem` file.

deploy/charts/kubernetes-cluster-prep/README.md
```diff
@@ -0,0 +1,129 @@
+# Helm chart for Conjur Kubernetes cluster preperation
```
diverdane, line 1: Typo (should be "Preparation", and probably should use upper case for the title:
```
# Helm Chart for Conjur Kubernetes Cluster Preparation
```

deploy/charts/kubernetes-cluster-prep/README.md
```diff
@@ -0,0 +1,129 @@
...
+
+This Helm chart is used to create Kubernetes objects that are required to 
+enable Conjur Kubernetes authentication
```
diverdane, line 4: Maybe change:
```
enable Conjur Kubernetes authentication
```
to
```
enable Conjur Kubernetes authentication on a Kubernetes cluster.
```

deploy/charts/kubernetes-cluster-prep/Chart.yaml
diverdane: https://stackoverflow.com/questions/58154658/where-is-a-helm-icon-url-used

diverdane (commented):
Still going through README.md

deploy/charts/kubernetes-cluster-prep/templates/clusterrole.yaml
rpothier: That's already there. Helm must add it.

deploy/charts/kubernetes-cluster-prep/templates/golden_configmap.yaml
rpothier: same as above

deploy/charts/kubernetes-cluster-prep/templates/golden_configmap.yaml
```diff
@@ -0,0 +1,42 @@
...
+      release: {{ .Release.Name }}
+      heritage: {{ .Release.Service }}
+      app.kubernetes.io/name: golden-configmap
+      app.kubernetes.io/component: "conjur-config"
+      app.kubernetes.io/instance: {{ .Release.Namespace }}
+      app.kubernetes.io/part-of: "authn-k8s-cluster-config"
+      helm.sh/chart: {{ .Chart.Name }}-{{ .Chart.Version }}
+data:
+    # authn-k8s Configuration 
+    authnK8sAuthenticatorID: {{ required "A valid authnK8s.authenticatorID is required!" .Values.authnK8s.authenticatorID }}
```
BradleyBoutcher, line 16: I think we should enforce required values with the schema validation rather than doing it within the charts. What do you think @diverdane ?

deploy/charts/kubernetes-cluster-prep/templates/golden_configmap.yaml
```diff
@@ -0,0 +1,42 @@
...
+    {{- if eq .Values.authnK8s.clusterRole.create true }}
+    authnK8sClusterRole: {{ .Values.authnK8s.clusterRole.name | default "authn-k8s-clusterrole" }}
```
BradleyBoutcher, line 18: If we set the values for `authnk8s.clusterRole.name` to "authn-k8s-clusterrole" in the `values.yml` file, do we need to check for a default value here? I guess my concern is if we change the default value, we have to go through and make sure each default check in each template uses the same value, whereas the `values.yml` file can act as the "defaults" implicitly. What do you think @diverdane ?

BradleyBoutcher: Additionally, our schema validation makes sure that the value for `clusterRole.name` is a value with more than one characters, so it's not possible for the user to pass in a null or empty value without getting an error message, which I think would be the correct response, rather than proceeding with a default value that's hidden from the user.

deploy/charts/kubernetes-cluster-prep/templates/golden_configmap.yaml
rpothier: yea, we don't need to have checks in two places. It does however add a bit of documentation to the user
to see that it is required if the user was browsing these files. If it is in both places, which one does the user see?

deploy/charts/kubernetes-cluster-prep/templates/golden_configmap.yaml
rpothier: So if it's empty, would it error before setting the default? So in that case there is no reason to set it as you mention.

deploy/charts/kubernetes-cluster-prep/templates/golden_configmap.yaml
diverdane: This is a bit involved...

There are some situations where we have "conditionally required" fields (how's that for an oxymoron?). For example, the `authnK8s.ServiceAccount.name` name is only required if `authnK8s.serviceAccount.create` is set to "false" (reason being: if the user is "bringing their own" ServiceAccount, they need to tell us what the name is, otherwise we can create one with a default name.

For the "conditionally required" fields, I think it's MUCH, MUCH easier to add the `required` inside the Helm manifest templates, since the Helm manifest templates have the logic all laid out, and the conditional logic is clear. Plus, I'm not sure if the `helm-unittest` plugin that we use recognized the conditional required fields that are defined in side the JSON schema (NOTE: `helm-unittest` does recognize unconditional required fields, but it does not recognize any or is not aware of the regex pattern checks that we add to schema validator).

So I think the "conditional required" fields need to go in the Helm manifest templates.

For the "unconditional required" fields, I think it comes down to this choice:
(1) If we add these to the schema validator, then the chart will not deploy ANY objects if the user is missing that required field.
(2) BUT, if we add these to the Helm manifest templates, it's easier to find all required fields since they're defined in one place.
(3) Defining them in both places wouldn't be too bad, IMO. A little non-DRY, but I think it's okay.

I have a slight preference for (2). WDYT?

deploy/charts/kubernetes-cluster-prep/templates/golden_configmap.yaml
diverdane: The reason this `default` shows up in the chart is because we want this functionality:
- If the user is "bringing their own" ClusterRole, then `authnK8s.clusterRole.create` is set to `false`. In this case, they need to tell us the ClusterRole name, so `authnK8s.clusterRole.name` is required.
- If the user is having the chart create a ClusterRole for them, then `authnK8s.clusterRole.create` is set to `true`. In this case, we can create a ClusterRole with a default name of `authn-k8s-clusterrole`, or they can tell us what to name it.

Checking this in schema would be a bit tricky.

deploy/charts/kubernetes-cluster-prep/templates/golden_configmap.yaml
rpothier: Since the purpose of this is to make things easier for the user then options 2 or 3 seem good. Option 3 would be good if there is further validation that can be done.

helm/kubernetes-cluster-prep/templates/serviceaccount.yaml
```diff
@@ -0,0 +1,14 @@
+{{- if .Values.authnK8s.serviceAccount.create -}}
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: {{ .Values.authnK8s.serviceAccount.name }}
```
diverdane, line 5: This needs to be:
```
 name: {{ .Values.authnK8s.serviceAccount.name | default "authn-k8s-serviceaccount" }}
```
This is another case where we want a "conditional default". In this case, we only want to provide a default when `Values.authnK8s.serviceAccount.create` is set to `true`. Since this default definition is made conditional by the `if .Values.authnK8s.serviceaccount.create` on Line 1.

Conversely, when the user "brings their own" ServiceAccount (`.Values.authnK8s.service.create` is `false`), then we DEFINITELY don't want to default the ServiceAccount name, because we need the user to tell us what their ServiceAccount name is that they want us to use. If we put defaults into the `values.yaml` for this setting, it would be an unconditional default.

diverdane (commented):
Hey Rob,

It's getting real close. I have a couple of recommended changes to values.yaml and some of the manifest tempates.

helm/kubernetes-cluster-prep/README.md
```diff
@@ -0,0 +1,288 @@
+# Helm Chart for Conjur Kubernetes Cluster Preparation
+
+## Table of Contents
+
+- [Helm Chart for Conjur Kubernetes Cluster Preparation](#helm-chart-for-conjur-kubernetes-cluster-preparation)
+  * [Table of Contents](#table-of-contents)
+  * [Overview](#overview)
+    + [Objects Created](#objects-created)
+  * [Preparing the Kubernetes Cluster for Conjur Authentication](#preparing-the-kubernetes-cluster-for-conjur-authentication)
+  * [Examples: Running Helm Install](#examples--running-helm-install)
```
diverdane, line 10: I think the double dashes `--` need to be replace with a single dash `-` in this line (line 10) and the next line. Otherwise, the link doesn't work.

helm/kubernetes-cluster-prep/values.yaml
```diff
@@ -0,0 +1,50 @@
...
+  # certificateFilePath: "files/conjur-cert.pem"
+  # certificateFilePath:
+
+  # Base64-encoded Conjur certificate file
+  # certificateBase64:
+
+authnK8s:
+  # Conjur authn-k8s authenticator ID to use for authentication.
+  # (This is a required value).
+  # authenticatorID:
```
diverdane, line 22: This can be uncommented, i.e. `  authenticatorID:`.  Sorry to keep going back and forth on this, but I know our UT will work fine with this uncommented, so better for user to make it visible.

diverdane (commented):
Looking AWESOME. A couple of nits, and it's ready.

helm/kubernetes-cluster-prep/README.md
rpothier: updated

helm/kubernetes-cluster-prep/values.yaml
rpothier: uncommented

diverdane (approved):
LGTM!!! Nice work!!!

Issue #252, opened 2021-03-19
Authenticator client sometimes panics when trying to kick off authentication


izgeri:
## Summary
The v0.19.1 authenticator client panics and fails for some applications. The customer reported that most applications were up and running fine, but one of them had this issue:
```
INFO: main.go:16: CAKC048 Kubernetes Authenticator Client v0.19.1-ee98afe starting up...
INFO: authenticator.go:207: CAKC040 Authenticating as user 'host/conjur/authn-k8s/ut-dev/apps/oa/stateful_set/hazelcast-node-analytic-data'
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x28 pc=0x7c2751]
```

> These ... are running hazelcast (https://hazelcast.com/) which makes me suspicious of something related to that app, but then again, it’s been working fine for months on 0.16.0. The deployments/pods have no memory limits set. I have tried a few things to troubleshoot including deployments with some versions of the authenticator between 0.16.0 and 0.19.1. This produced some interesting results:
> 
> - 0.16.0, 0.16.1: Works great without a problem
> - 0.18.1, 1.19.0, 1.19.1: All of these versions fail with the above errors.

## Steps to Reproduce
TBD - so far we have not been able to reproduce this locally

## Expected Results
The authenticator client delivers an access token as expected.

## Actual Results (including error logs, if applicable)
The authenticator client panics and fails.

More detailed logs:
```
NFO:  2021/03/17 19:55:01.166872 main.go:16: CAKC048 Kubernetes Authenticator Client v0.19.1-ee98afe starting up...
INFO:  2021/03/17 19:55:01.253013 authenticator.go:207: CAKC040 Authenticating as user 'host/conjur/authn-k8s/ut-dev/apps/oa/stateful_set/hazelcast-node-analytic-data'
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x28 pc=0x7c2751]

goroutine 1 [running]:
github.com/cyberark/conjur-authn-k8s-client/pkg/utils.VerifyFileExists(0x84dc32, 0x1a, 0x0, 0xc00043e000)
	/opt/conjur-authn-k8s-client/pkg/utils/file.go:50 +0xb1
github.com/cyberark/conjur-authn-k8s-client/pkg/utils.WaitForFile.func1(0xc00009a060, 0xc000092080)
[...]

PR #256, opened 2021-03-25
Fixes bug in error handling within the `VerifyFileExists` method

doodlesbykumbi:
### What does this PR do?

Fixes bug in error handling within the `VerifyFileExists` method that resulted in a panic when the error from `os.Stat` was not `ErrNotExist`. The fix includes introducing the `CAKC058` error and log message for when the path to a file exists but is not a regular file.

Resolves #https://github.com/cyberark/conjur-authn-k8s-client/issues/252

#### Change log
- [x] The CHANGELOG has been updated, or

#### Test coverage
- [x] This PR includes new unit and integration tests to go with the code changes, or

#### Documentation

#### Manual tests


diverdane (commented):
I think we really should add checking for and logging of **permissions errors** (in addition to logging errors that indicate non-regular files). There's a few reasons why I'm thinking that we're getting hit with permissions errors when checking for the existence of the authenticator client certificate:

- If you look at the first (top) entry in the crash traceback:

  ```
   github.com/cyberark/conjur-authn-k8s-client/pkg/utils.VerifyFileExists(0x84dc32, 0x1a, 0x0, 0xc00043e000)
   	/opt/conjur-authn-k8s-client/pkg/utils/file.go:50 +0xb1
   ```
   the first 2 hexadecimal arguments shown on the stack, i.e. `0x84dc32, 0x1a`, represent a *pointer* and a *length* for the `path` string argument being passed to the function. (The last two represent the return value, which also happened to get added to the stack when Golang gets compiled/run). 
   
   This says that the length of the string `path` being passed is `0x1a`, or 26 decimal. This is EXACTLY what we're expecting in this case, because we're checking for the authenticator client cert at the path `/etc/conjur/ssl/client.pem`, which is 26 characters long. So it appears that at least the length of the string is not junk... which leads me to believe that the pointer pointing to the text is probably not junk either.
   
* The applications are able to authenticate when version 0.16.1 of the authenticator was used, but fail to authenticate when 0.19.1 is used.

   One of the differences between v0.16.1 and v0.19.1 that might affect access to the cert at `/etc/conjur/ssl/client.pem` is that the Dockerfile for building the authenticator was changed a bit between these versions:
       - For v0.16.1, the `/etc/conjur/ssl` directory is created with permissions `0777`, and that directory is created by user `root` in the Docker build process.
       - For v0.19.1, the `/etc/conjur/ssl` directory is created with permissions `0770`, and that directory is created by user `authenticator` (and the container's user is set to `authenticator`).
       
  Everything looks like it *should* be okay and the `/etc/conjur/ssl/client.pem` should be readable in both cases, but it does seem a bit suspicious.
  
  I started to add this check in, and it was pretty easy in the functional/production code, but the UT for this is difficult as it's hard to trigger a permissions error in calls to `os.Stat()`. So this needs to be mocked. I'm close to having something that might work.

diverdane:
@doodlesbykumbi I've added error logging for a file permissions error (for reasons listed earlier), but unit testing this required a restructuring of the mocking that was done for testing files.go.

Let me know what you think... if this looks like a bad idea you can do a force push to remove this change.

diverdane (dismissed):
Looks fantastic. Mocking the way it should be done. :)
I had some minor suggestions to consider, but I don't have a preference on these one way or another, so your choice.

CHANGELOG.md
```diff
@@ -5,6 +5,13 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Fixed
+- Fixes bug in error handling within the `VerifyFileExists` method that resulted in a
+  panic when the error from `os.Stat` was not `ErrNotExist`. The fix includes introducing
+  the `CAKC058` error and log message for a file permissions error and a the
```
diverdane, line 11: Dangit. It should be `error and the` and not `error and a the`.

pkg/log/log_messages.go
```diff
@@ -70,3 +70,5 @@ const CAKC054 string = "CAKC054 Failed to delete file %s"
 const CAKC055 string = "CAKC055 Cert placement failed with the following error:\n%s"
 const CAKC056 string = "CAKC056 File %s does not exist"
 const CAKC057 string = "CAKC057 Removing file %s"
+const CAKC058 string = "CAKC058 Permissions error when checking if file exists: %s"
+const CAKC059 string = "CAKC059 Path exists but does not contain regular file: %s"
```
orenbm, line 74: @shulifink can you review these messages please?

diverdane: @shulifink , any chance you can review these log messages? Thanks!

shulifink:
@suefran  - can you please take this?

suefran:
I'm on it, thanks!

pkg/log/log_messages.go
```diff
@@ -70,3 +70,5 @@ const CAKC054 string = "CAKC054 Failed to delete file %s"
 const CAKC055 string = "CAKC055 Cert placement failed with the following error:\n%s"
 const CAKC056 string = "CAKC056 File %s does not exist"
 const CAKC057 string = "CAKC057 Removing file %s"
+const CAKC058 string = "CAKC058 Permissions error when checking if file exists: %s"
```
suefran, line 73: Suggested wording:

"CAKC058 Permissions error occurred when checking if file exists: %s"

pkg/log/log_messages.go
suefran, line 74: Suggested wording:

"CAKC059 Path exists but does not contain expected file: %s"

(If "expected file" skews the meaning, feel free to ignore the suggestion. I suggested this as "regular file" was too non-specific - if it isn't regular, then is it irregular? I didn't think that was what you intended.)

suefran (dismissed):

diverdane (approved):
LGTM!!!

PR #279, opened 2021-04-08
Updating get-conjur-cert.sh

### LINKED ISSUE #278: The script get-conjur-cert handles relative file paths
## Summary
There are a few issues with the get-conjur-cert.sh

1. With the -f option if the path is relative, it is places relative to the bin directory, not the users directory.
2. For the external case, it would be nice to check if openssl was installed to tell the user why it's failing.
3. Depending on the kubectl version, kubectl run might make a depoyment or pod, it should always be deployment.
4. Verify should work if there is a http:// on the URL or not.

## Steps to Reproduce

Running get-conjur-cert.sh

## Expected Results

The script should work in the above scenarios.

## Actual Results (including error logs, if applicable)
A clear and concise description of what actually did happen.

## Reproducible
   * [x ] Always 
   * [ ] Sometimes
   * [ ] Non-Reproducible
   
## Version/Tag number
Latest

rpothier:
There are a few issues with the get-conjur-cert.sh that need to be fixed.

- When using the -f option if the path is relative, the file destination is relative
  to the script directory, not the users directory.
- For the external case, it would be nice to check if openssl was installed to tell
  the user why it's failing.
- Depending on the kubectl version, kubectl run might make a depoyment or pod,
  it should always be deployment.
- Verify should work if there is a http:// prefix on the URL or not.

### What does this PR do?

Fixes the above issues.

Resolves #278 

#### Change log

#### Test coverage

#### Documentation
- [x] Docs (e.g. `README`s) were updated in this PR, and/or there is a follow-on issue to update docs, or

#### Manual tests

### DIFF
--- a/helm/kubernetes-cluster-prep/bin/get-conjur-cert.sh
+++ b/helm/kubernetes-cluster-prep/bin/get-conjur-cert.sh
@@ -17,7 +17,8 @@ set -euo pipefail
 # instance that is internal to the Kubernetes cluster:
 #     get-conjur-cert.sh -v -i -u https://conjur.conjur-ns.svc.cluster.local
  
-# Execute from the directory in which this script exists
+# Save the users directory and execute from the directory in which this script exists
+user_dir=$(pwd)
 cd "$(dirname "$0")"
 
 # Default destination filepath for certificate file and default test
@@ -89,17 +90,17 @@ function main() {
 
   # Process destination filepath argument
   if [[ -n "$filepath" ]]; then
-    if [[ "$filepath" = "/*" ]]; then
+    if [[ "${filepath:0:1}" = "/" ]]; then
       # Absolute filepath
       cert_filepath="$filepath"
     else
       # Relative filepath
-      cert_filepath="$(pwd)/$filepath"
+      cert_filepath="$user_dir/$filepath"
     fi
     echo "Saving certificate to $cert_filepath"
   else
     # User's perspective file path
-    echo "Saving certificate to $(pwd)/$cert_filepath"
+    echo "Saving certificate to default path of $(pwd)/$cert_filepath"
   fi
 
   domain_name="$(get_domain_name $conjur_url)"
@@ -109,6 +110,14 @@ function main() {
   # Read the certificate from either internal or external Conjur instance
   if [[ "$internal_addr" = false ]]; then
     echo "Assuming Conjur instance is outside of the Kubernetes cluster."
+
+    status=0
+    openssl_ver="$(openssl version)" || status="$?"
+    if [ "$status" != 0 ] ; then
+      echo "OpenSSL is required to gather the SSL certificate"
+      exit 1
+    fi
+    echo "Using " "$openssl_ver"
 
     # Retrieve certificate
     cert="$(sh -c "$ssl_cmd")"
@@ -134,7 +143,7 @@ function main() {
 
     # Verify if desired
     if [ "$verify" = true ] ; then
-      k8s_verify_certificate "$openssl_pod" "$cert_filepath" "$conjur_url"
+      k8s_verify_certificate "$openssl_pod" "$cert_filepath" "$domain_name"
     fi
 
     # Delete the openssl test deployment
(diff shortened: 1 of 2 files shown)


helm/kubernetes-cluster-prep/bin/get-conjur-cert.sh
```diff
@@ -110,6 +111,14 @@ function main() {
   if [[ "$internal_addr" = false ]]; then
     echo "Assuming Conjur instance is outside of the Kubernetes cluster."
 
+    openssl_ver="$(openssl version)"
```
diverdane, line 114: If you combine this line with the next with a pipeline logical OR:
```
    status=0
    openssl_ver="$(openssl version) || status="$?"
    if [ "$status" != 0 ] ; then
      echo "OpenSSL is required to gather the SSL certificate"
      exit 1
    fi
    echo "Using " "$openssl_ver"
```
then we'd be able to use the `set -e` flag for the script.

helm/kubernetes-cluster-prep/bin/get-conjur-cert.sh
```diff
@@ -110,6 +111,14 @@ function main() {
...
+    status="$?"
+    if [ "$status" != 0 ] ; then
+      echo "OpenSSL is required to gather the SSL certificate"
+      exit
```
diverdane, line 118: Should be `exit 1`

helm/kubernetes-cluster-prep/bin/get-conjur-cert.sh
```diff
@@ -181,9 +190,9 @@ function ensure_openssl_pod_created() {
     # Create a test deployment if it hasn't been created already
     openssl_pod="$(get_openssl_pod $openssl_deployment)"
     if [ -z "$openssl_pod" ]; then
-        kubectl run "$openssl_deployment" \
+        kubectl create deployment "$openssl_deployment" \
            --image cyberark/conjur-cli:5 \
-           --command sleep infinity
+           -- command sleep infinity
```
diverdane, line 195: Can you double-check this, it looks like `command` might not be needed here? I could be wrong.

helm/kubernetes-cluster-prep/bin/get-conjur-cert.sh
```diff
@@ -205,14 +214,16 @@ function k8s_verify_certificate() {
   cert_filepath="$2"
   conjur_url="$3"
```
diverdane, line 215: We should change this passed argument from `conjur_url` to `domain_name`. You'd also have to change where `k8s_verify_certificate` gets invoked earlier to actually pass the `domain_name` variable, but that variable is available at that invocation.

This would simplify and make things clearer in that you'd be able to skip the next line where you call `get_domain_name`, and use the domain name directly, e.g.:
```
  curl_cmd="curl --cacert $cert_filename $domain_name"
```

Or if you'd prefer, change the next line to:
```
  url="https://$domain_name"
```

diverdane (commented):
Nice cleanup, this is looking robust!
A couple of suggested tweaks.

helm/kubernetes-cluster-prep/bin/get-conjur-cert.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-set -euo pipefail
+set -uo pipefail
```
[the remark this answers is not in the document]
rpothier: yes, will check below.

helm/kubernetes-cluster-prep/bin/get-conjur-cert.sh
rpothier: ok, that wrks, thanks!

helm/kubernetes-cluster-prep/bin/get-conjur-cert.sh
rpothier: yes, it's not needed. I removed it.

helm/kubernetes-cluster-prep/bin/get-conjur-cert.sh
rpothier: yes, that is cleaner

diverdane (approved):
Nice! LGTM!

PR #281, opened 2021-04-09
Adds mocking for lookup function in app namespace prep chart

### LINKED ISSUE #270: Add mocking for lookup function to Namespace prep chart
## Is your feature request related to a problem? Please describe.

The current implementation of the Namespace prep Helm chart uses lookup inside the chart as follows:
Lookup of a dictionary of the COnfigmap data:

```
{{- $g := (lookup "v1" "ConfigMap" .Values.authnk8s.namespace .Values.authnk8s.goldenConfigmap).data -}}
```

And then this dictionary gets used later:

```
- kind: ServiceAccount
  name: {{ get $g "authnK8sServiceAccount" }}
```

However, the lookup function cannot be used in Helm unittest or Helm lint.

In order to be able to run Helm unittest and Helm lint on the Namespace Helm chart, the instances of the Helm lookup() function will need to be mocked. This can be done by adding new functions/methods in `_helpers.tpl` to do this optional mocking for every location in the Helm chart templates that refer to golden configmap values. The pseudo code for each lookup would be something like this:

```
{{/*
Return the authnK8sServiceAccount setting, which can either be
the result of a run-time lookup in the Golden ConfigMap, or a mocked value
for testing
*/}}
{{- define "namespace-prep.authn-k8s-serviceaccount" -}}
{{- if .Values.test.mock.enable -}}
{{- default "my-serviceaccount" .Values.test.mock.serviceAccount" -}}
{{- else }}
{{- (lookup "v1" "ConfigMap" .Values.authnk8s.namespace .Values.authnk8s.goldenConfigmap).data.authnK8sServiceAccount -}}
{{- end }}
{{- end }}
```

If defining a method in _helpers.tpl doesn't work, then another way to do this would be to add this  conditional logic directly in the Helm manifest template. This approach should definitely work, but the downside would be that our templates would be very verbose with all of this conditional logic.

For example, this will DEFINITELY work as a (verbose) fallback (I've tested this):
```
- kind: ServiceAccount
  {{- if .Values.test.mock.enable }}
  name: {{ .Values.test.mock.authnK8sServiceAccount }}
  {{- else }}
  name: {{ (lookup "v1" "ConfigMap" (print .Values.authnk8s.namespace) (print .Values.authnk8s.goldenConfigMap)).data.authnK8sServiceAccount }}
  {{- end }}
```
 
## Describe the solution you would like

The Namespace prep Helm chart should allow the lookups to be mocked. This can be implemented either:
- Creating functions in _helpers.tpl as described above
- OR.... (less desirable because of verbosity in manifest templates) Adding the condtional logic directly in the Helm chart manifest templates that
   chooses between lookup vs. using mock values from .Values.test

## Describe alternatives you have considered

Alternative is to do all of the testing that we want to do by using a local Kubernetes KinD cluster, and developing tests that run `Helm install` with various chart settings and verifies results are as expected.

## DoD
-[ ] Namespace prep chart supports mocking of values that normally use the Helm lookup() function

diverdane:
### What does this PR do?
This change does the following for the application Namespace prep Helm chart:
- Adds mocking for the Helm `lookup()` function that is used in manifest templates.
- Moves the `NOTES.txt` file to `templates/` directory.
- Adds a `test-lint` script.

Resolves #270

#### Change log

#### Test coverage

#### Documentation

#### Manual tests


rpothier (approved):
LGTM!

PR #295, opened 2021-05-04
Adding Helm Test to kubernetes-cluster-prep

### LINKED ISSUE #229: There is an initial framework for the cluster helm test
# Overview
After a Kubernetes administrator has run `helm install ...` or `helm upgrade ...` using the cluster prep Helm chart, we would like to provide a way for the admin to validate the Helm release, i.e. to prove that the Kubernetes objects that have been deployed can be trusted to support the deployment of applications that will use the associated authn-k8s authenticator.

This will require the implementation of a Helm test (see https://helm.sh/docs/topics/chart_tests/) that can be run on-demand by the Kubernetes administrator after `helm install ...` and `helm upgrade ...`.

Helm tests typically use the deployment of Kubernetes Pods or Jobs to run custom test applications to exercise the functionality of the Release that they're designed to test.

For the cluster prep Helm chart, the Helm test will refer to the data in the Golden ConfigMap (using a Pod volume mount), and will use `curl` and `openssl` to access the configured Conjur instance (based on Conjur URL) to validate that the configuration is correct for this Conjur instance.

For an example of a Helm test as a reference, see:
    https://github.com/cyberark/conjur-oss-helm-chart/tree/master/conjur-oss/templates/tests

# Tests to be Performed

__NOTE: This user story covers creating just the basic framework for the cluster prep Helm chart.
   More tests will be added subsequently via Issue #230, and Issue #231.__

For the initial implementation of the Helm test for the Kubernetes cluster prep Helm chart,
we'll include only a couple of simple tests (the Golden ConfigMap will be volume-mounted to the Helm test Pod/Job):

- `curl -k <Conjur URL>`
    This verifies that the Conjur server can be reached (i.e. logical `ping`) using the Conjur URL,
    regardless of whether the Conjur SSL certificate is accurate or not.
- Use the `openssl` utility to retrieve the Conjur SSL certificate using the configured Conjur URL,
   and verify that the retrieved SSL certificate matches the configured Conjur SSL certificate.
- OPTIONAL: Add more SSL certificate checking using openssl???

# Required Components for Helm Test Framework

## Dockerfile
The Helm test will require a custom image that contains:
- `curl`
- `openssl`
Here's a rough idea of what's needed:
```
FROM alpine:3.12

RUN mkdir -p /tests
WORKDIR /tests

# Install Docker client
RUN apk add --no-cache curl openssl bash

ENTRYPOINT [ "/tests/test" ]
```

## build script
We'll need a build script to build a Docker image via the Dockerfile described above

## Manifest for test Pod/Job
A Kubernetes manifest for a Pod or Job to run the test scripts is needed.
Here is a rough idea of what's needed, based on an early P.O.C.:
```
apiVersion: v1
kind: Pod
metadata:
  name: {{ .Release.Name }}-cluster-prep-test
  labels:
  annotations:
    "helm.sh/hook": test-success
spec:
  initContainers:
  - name: {{ .Release.Name }}-bats-init
    image: bats/bats:v1.1
    command:
    - "bash"
    - "-exc"
    - |
      # copy bats to tools dir
      cp -R /opt/bats/libexec/bats-core/ /tools/bats/
    volumeMounts:
    - mountPath: /tools
      name: tools
  containers:
  - name: {{ .Release.Name }}-test
    image: diverdane/conjur-cluster-prep-test:latest
    workingDir: "/tools/bats"
    command: ["/tools/bats/bats", "-t", "/tests/run.sh"]
    env:
    - name: PATH
      value: "/tools/bats:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"
    envFrom: 
    - configMapRef: 
        name: authn-k8s-config-map
    volumeMounts:
    - mountPath: /tests
      name: tests
      readOnly: true
    - mountPath: /tools
      name: tools
  volumes:
  - name: tests
    configMap:
      name: {{ .Release.Name }}-tests-configmap
  - name: tools
    emptyDir: {}
  restartPolicy: Never
```

## OPTIONAL: Manifest for test ConfigMap containing 'bats' test scripts
Optionally, the tests scripts can be provided to the test Pod/Job as a ConfigMap,
similar to what's done for the Conjur OSS Helm chart Helm test.

Here's a rough idea of what's needed for the first `curl -k ...` test, based on an early P.O.C.:
```
apiVersion: v1
kind: ConfigMap
metadata:
  name: {{ .Release.Name }}-tests-configmap
data:
  run.sh: |-
    @test "Testing basic connectivity to Conjur" {
        curl -k "$conjurApplianceUrl:443"
    }
```

# Test Results Visiblity
__The test results for failed test cases must be visible on the `helm test` command line.
   When a failure occurs, it should be clear to the person running the test what the specific failure was.
   It might be sufficient to dump the test Pod/Job logs when failure occurs.__

# DoD
- Helm test components created:
   - Dockerfile
   - Build/push script
   - Test Pod/Job manifest
   - Optional: Test ConfigMap
- Helm test happy path works
- Helm test fails for invalid Conjur config
- Failures are clearly described in `helm test ...` command output
- Test image is pushed for builds on master branch using version as a tag.

rpothier:
### What does this PR do?
This adds the  Helm Test to kubernetes-cluster-prep

Resolves #229 

#### Change log

#### Test coverage

#### Documentation

#### Manual tests

### DIFF
--- a/helm/kubernetes-cluster-prep/templates/tests/_helm_test.bats.txt
+++ b/helm/kubernetes-cluster-prep/templates/tests/_helm_test.bats.txt
@@ -1,0 +1,31 @@
+# Define a Helm named template so that this BATS script can be included
+# in a Helm test ConfigMap. This can be left commented out; Helm will
+# still be able to find this template, and this file can be kept as
+# valid BATS syntax.
+#
+{{- define "kube-cluster-prep.helm-test.bats" }}
+#!/usr/bin/env bats
+
+# Run out of same directory in which this script resides
+cd "/tests"
+source "./utils.sh"
+
+source "/bats/bats-support/load.bash"
+source "/bats/bats-assert/load.bash"
+source "/bats/bats-file/load.bash"
+
+# Baseline BATS test result color
+text_color "$MAGENTA"
+
+@test "Conjur Appliance URL is a reachable address" {
+  cmd=(curl -k --connect-timeout 5 "$conjurApplianceUrl":443)
+  display_info "Running ${cmd[@]}"
+  run "${cmd[@]}"
+  if [ "$status" -ne 0 ]; then
+    display_error "Please check configured Conjur Appliance URL.\n" \
+                  "It is not reachable via 'curl -k'"
+  fi
+  assert_success
+}
+
+{{- end }}
--- a/helm/kubernetes-cluster-prep/test-helm
+++ b/helm/kubernetes-cluster-prep/test-helm
@@ -1,0 +1,71 @@
+#!/bin/bash
+#
+# Usage:
+#     ./test-helm [<Helm test release>]
+#
+# Example:
+#     ./test-helm my-test-release
+
+print_usage() {
+  echo "Usage:"
+  echo "    This script will test the Helm release"
+  echo ""
+  echo "Syntax:"
+  echo "    $0 [Options]"
+  echo "    Options:"
+  echo "    -h           Show help"
+  echo "    -i           Install the Helm release"
+  echo "    -u           Uninstall the Helm release"
+  echo "    -r           Specify the Helm release"
+  echo "    -None at this time"
+}
+
+helm_release="helm-test-release"
+install=false
+uninstall=false
+conjur_account="${CONJUR_ACCOUNT:-myConjurAccount}"
+conjur_appliance_url="${CONJUR_APPLIANCE_URL:-https://conjur-oss.conjur-oss.svc.cluster.local}"
+conjur_certificate_file_path="${CONJUR_CERT_FILE_PATH:-files/conjur-cert.pem}"
+authenticator_id="${AUTHENTICATOR_ID:-my-authenticator-id}"
+
+function main() {
+  # Process command line options
+  local OPTIND
+  while getopts ':h:ir:u' flag; do
+    case "${flag}" in
+      #a) conjur_account=${OPTARG} ;;
+      h) print_usage; exit 0 ;;
+      i) install=true ;;
+      r) helm_release=${OPTARG} ;;
+      u) uninstall=true ;;
+      *) echo "Invalid argument -${OPTARG}" >&2; echo; print_usage ; exit 1;;
+    esac
+  done
+  shift $((OPTIND-1))
+
+  if [ "$uninstall" = true ] ; then
+    echo "Cleaning up previous release, if necessary"
+    helm status "$helm_release" &> /dev/null 
+    status="$?"
+    echo "status=" "$status"
+
+    if [ "$status" = 0 ] ; then
+      echo "Deleting"
+      helm delete "$helm_release"
+    fi
+    echo "Done cleaning up previous release"
+    sleep 6
+  fi
+  if [ "$install" = true ] ; then
+    echo "Helm installing a new release" "$conjur_account"
+    helm install "$helm_release" . \
+         --set conjur.account="$conjur_account" \
+         --set conjur.applianceUrl="$conjur_appliance_url" \
+         --set conjur.certificateFilePath="$conjur_certificate_file_path" \
+         --set authnK8s.authenticatorID="$authenticator_id"
+  fi
+  echo "Running Helm test"
+  helm test "$helm_release" --logs
+}
+
+main "$@"
--- a/helm/kubernetes-cluster-prep/.helmignore
+++ b/helm/kubernetes-cluster-prep/.helmignore
@@ -21,3 +21,6 @@
 .idea/
 *.tmproj
 .vscode/
+# Helm test image Dockerfile and test utilities
+templates/tests/Dockerfile
+templates/tests/build.sh
--- a/helm/kubernetes-cluster-prep/bin/get-conjur-cert.sh
+++ b/helm/kubernetes-cluster-prep/bin/get-conjur-cert.sh
@@ -178,6 +178,12 @@ function get_domain_name() {
     echo "$1" | sed -e 's|^[^/]*//||' -e 's|/.*$||'
 }
 
+function get_openssl_deployment() {
+    openssl_deployment="$1"
+
+    kubectl get pod -l "app=$openssl_deployment" -o jsonpath='{.items[*].metadata.name}'
+}
+
 function get_openssl_pod() {
     openssl_deployment="$1"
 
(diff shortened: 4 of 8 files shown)


helm/kubernetes-cluster-prep/templates/tests/_helm_test.bats
```diff
@@ -0,0 +1,33 @@
...
+
+source "/bats/bats-support/load.bash"
+source "/bats/bats-assert/load.bash"
+source "/bats/bats-file/load.bash"
+
+# Baseline BATS test result color
+text_color "$MAGENTA"
+
+@test "Conjur Appliance URL is a reachable address" {
+  status=0
```
diverdane, line 21: Now that we're loading `bats-assert`, we can use that functionality to condense this test case a little:
```
@test "Conjur Appliance URL is a reachable address" {
  cmd=(curl -k --connect-timeout 5 "$conjurApplianceUrl":443)
  display_info "Running ${cmd[@]}"
  run "${cmd[@]}"
  if [ "$status" -ne 0 ]; then
    display_error "Please check configured Conjur Appliance URL.\n" \
                  "It is not reachable via 'curl -k'"
  fi
  assert_success
}

File: helm/kubernetes-cluster-prep/templates/tests/build.sh [hunk 2]
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash -e
+
+# Run from dir in which this script resides
+cd "$(dirname "$0")"
+
+TAG=latest
```
diverdane, line 6: Do we need `TAG` to be settable via environment variable? e.g.:
```
TAG="${TAG:-latest}"
```
Or are we tagging the image that gets built (e.g. with "edge") outside of this script?

helm/kubernetes-cluster-prep/templates/tests/build.sh
```diff
@@ -0,0 +1,12 @@
...
+
+echo "================================================"
+echo "     Building Cluster Prep Helm test image"
+echo "================================================"
+docker build -t "rpothier/conjur-cluster-prep-test:$TAG" .
```
diverdane, line 11: Should be `cyberark/conjur-cluster-prep-test:$TAG`

helm/kubernetes-cluster-prep/test-helm
```diff
@@ -0,0 +1,72 @@
...
+print_usage() {
+  echo "Usage:"
+  echo "    This script will test the Helm release"
+  echo ""
+  echo "Syntax:"
+  echo "    $0 [Options]"
+  echo "    Options:"
+  echo "    -h           Show help"
+  echo "    -i           Install the Helm release"
+  echo "    -d           Unitall the Helm release"
```
diverdane, line 18: Typo `Unitall`

helm/kubernetes-cluster-prep/test-helm
```diff
@@ -0,0 +1,72 @@
...
+  # Process command line options
+  local OPTIND
+  while getopts ':d:h:ir:u' flag; do
+    case "${flag}" in
+      #a) conjur_account=${OPTARG} ;;
+      h) print_usage; exit 0 ;;
+      i) install=true ;;
+      r) helm_release=${OPTARG} ;;
+      d) uninstall=true ;;
+      u) uninstall=true ;;
```
diverdane, line 41: We have both `u` and `d` options, maybe `u` would be clearer?

diverdane (commented):
Looks good. A few minor comments.

helm/kubernetes-cluster-prep/templates/tests/_utils.sh
```diff
@@ -0,0 +1,65 @@
...
+  echo -e '\033[0;'"$color_code"'m'
+  #{{- else }}
+  echo
+  #{{- end }}
+}
+
+function banner() {
+  text_color "$1"
+  echo =====================================================================
+  echo -e "${@:2}"
```
diverdane, line 35: This is probably a personal style preference, but I saw a script that uses bash `shift` to sort of pop through the argument stack. This would avoid having to segment/slice the `$@` array (`${@:2}`).

Using `shift`, it would be:
```
function banner() {
  text_color "$1"
  shift
  echo =====================================================================
  echo -e "${@}"
  echo =====================================================================
  text_color "$RESET_COLOR"
}
```
Your choice.

helm/kubernetes-cluster-prep/test-helm
rpothier: Agreed

helm/kubernetes-cluster-prep/templates/tests/build.sh
rpothier: This file is now just a helper file, as the build happens in bin/build
https://github.com/cyberark/conjur-authn-k8s-client/blob/master/bin/build#L27
I can remove this file

helm/kubernetes-cluster-prep/templates/tests/build.sh
rpothier: same as above

helm/kubernetes-cluster-prep/templates/tests/_utils.sh
rpothier: I'll go with the shift, seems a little cleaner.

helm/kubernetes-cluster-prep/templates/tests/_helm_test.bats
rpothier: ok, added it.
(I had to add an assert_failure after the display_error.)

helm/kubernetes-cluster-prep/templates/tests/_helm_test.bats
```diff
@@ -0,0 +1,33 @@
+# Define a Helm named template so that this BATS script can be included
+# in a Helm test ConfigMap. This can be left commented out; Helm will
+# still be able to find this template, and this file can be kept as
+# valid BATS syntax.
+#
+#{{- define "kube-cluster-prep.helm-test.bats" }}
```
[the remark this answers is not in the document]
rpothier: hmm. on the fence on this. Lets try it out.

helm/kubernetes-cluster-prep/templates/tests/_helm_test.bats.txt
```diff
@@ -0,0 +1,33 @@
...
+
+source "/bats/bats-support/load.bash"
+source "/bats/bats-assert/load.bash"
+source "/bats/bats-file/load.bash"
+
+# Baseline BATS test result color
+text_color "$MAGENTA"
+
+@test "Conjur Appliance URL is a reachable address" {
+  status=0
```
diverdane, line 21: We can leave out the initialization of the `status` variable. The `run` command always declares/initializes that for us automatically.

I should have added the documentation for this earlier:
    https://bats-core.readthedocs.io/en/latest/writing-tests.html

Based on that documentation ^^^ (and the example it provides), the `run` command gives us these three variables automagically:
- $status
- $output
- $lines

helm/kubernetes-cluster-prep/templates/tests/_helm_test.bats.txt
```diff
@@ -0,0 +1,33 @@
...
+  cmd=(curl -k --connect-timeout 5 "$conjurApplianceUrl":443)
+  display_info "Running ${cmd[@]}"
+  run "${cmd[@]}"
+  if [ "$status" -ne 0 ]; then
+    display_error "Please check configured Conjur Appliance URL.\n" \
+                  "It is not reachable via 'curl -k'"
+    assert_failure
```
diverdane, line 28: We can leave out the `assert_failure` here, and just use the one `assert_success` that's outside of this conditional.

Some background: The `assert_failure` and `assert_success` basically assert on the value of the `$status` variable, so it's convenient to just add an `assert_success` at the very end of the test case.

diverdane (commented):
Looking good!
Just a couple of very minor tweaks, and it's good-to-go IMO!

diverdane (approved):
AWESOME SAUCE!

PR #300, opened 2021-05-13
Adding checks for Conjur account and authenticatorID

### LINKED ISSUE #230: The cluster helm test queries the /info endpoint for validation
# Overview
__This issue builds on the framework created in Issue #229. For an overview of the cluster prep
   Helm chart Helm test framework, refer to Issue #229.__

This issue adds a validation of the following Conjur configuration fields in the Golden ConfigMap:
- conjurAccount
- authnK8sAuthenticatorID
using the `/info` Conjur API endpoint. Since this endpoint is currently only available for
Conjur Enterprise, this feature will only work with Conjur Enterprise instances.

# Addition of `conjur.type` (can be "oss" or "enterprise") to chart values
A `conjur.type` field should be added to the chart's `values.yaml` file.
The default setting for this value will be "oss".

# Expecting `/info` access vs. "best effort"
- When `conjur.type` is explicitly set to "enterprise", this will signify that the `/info`
   endpoint is expected to work, and a lack of response is a failure.
- __When `conjur.type` is not explicitly set to "enterprise", then access to `/info`
   endpoint will be "best effort".__ That is, an attempt will be made to to access the
   `/info` endpoint. Failure to access this endpoint will not be considered a
   test failure.

# Parsing `/info` endpoint data and validating
When access to the `/info` endpoint is successful, the returned data
will be parsed for:
- Conjur account
- Enabled authn-k8s authenticator IDs

If the actual Conjur account does not match the `conjurAccount` value in the Golden ConfigMap,
then a test failure is declared.

If the enabled authn-k8s authenticator IDs do not contain the `authnK8sAuthenticatorID`
value in the Golden ConfigMap, then a test failure is declared.

# Test results
Test failures must appear in the command line output of the `helm test ...` command.
The output should be clear about which value has been misconfigured.

# Test environments
Testing must be done both with Conjur OSS instance and a Conjur Enterprise instance.

# DoD
- `conjur.type` added to chart values, defaulting to "oss"
- A JSON schema validation is added to `values.schema.json` for this new field
- Test cases added to cluster prep Helm chart test
- Testing performed on both Conjur OSS and Conjur Enterprise setup
- Helm test passes for "happy path" valid values
- Helm test fails for incorrect conjurAccount, and failure is clear on command line output
- Helm test fails for incorrect authnK8sAuthenticatorID, and failure is clear on command line output

rpothier:
Adding checks for Conjur account and authenticatorID in the Helm tests.

### What does this PR do?
This adds two Helm tests to validate the Conjur account and authenticatorID
that is retrieved from the /info endpoint.
The /info endpoint only works on the enterprise edition to extra checks were made
for the failing case if it wasn't enterprise.

Opted not to add new item to values.yaml
Instead of a setting of enterprise or oss, the test will parse the info endpoint.
If the data matches, it passes no matter what edition.
If the first test fails, then we do another check of the information gathered
from the curl of the endpoint. If it was oss, the results will not have lines
like account and name, it will just have "Authorization missing".

So for OSS, the tests will be skipped, as below.
```
ok 1 Conjur Appliance URL is a reachable address

ok 2 Conjur Account is valid # skip test due to the info endpoint is not available for OSS

ok 3 Conjur Authenticator is valid # skip test due to the info endpoint is not available for OSS

```
Resolves #230 

#### Change log

#### Test coverage

#### Documentation
- [x] Docs (e.g. `README`s) were updated in this PR, and/or there is a follow-on issue to update docs, or

#### Manual tests

### DIFF
--- a/helm/kubernetes-cluster-prep/templates/tests/_helm_test.bats.txt
+++ b/helm/kubernetes-cluster-prep/templates/tests/_helm_test.bats.txt
@@ -16,6 +16,7 @@ source "/bats/bats-file/load.bash"
 
 readonly ACCESS_TOKEN_FILE="/run/conjur/access-token"
 readonly AUTHN_LOG_FILE="/run/conjur/authn-logs.txt"
+readonly TEMP_INFO_FILE="/info.txt"
 readonly AUTHN_TIMEOUT_SECS=5
 
 # Baseline BATS test result color
@@ -74,4 +75,43 @@ text_color "$MAGENTA"
   assert_failure
 }
 {{- end }}
+
+@test "Conjur Account is valid" {
+  display_info "Validating account from the info endpoint for enterprise"
+
+  cmd=( curl -k --connect-timeout 5 "$conjurApplianceUrl"/info -o "$TEMP_INFO_FILE")
+  display_info "Running ${cmd[@]}"
+  "${cmd[@]}"
+  run grep account.*"$conjurAccount" "$TEMP_INFO_FILE"
+
+  if [ "$status" -ne 0 ]; then
+    if (grep account "$TEMP_INFO_FILE"); then
+        display_error "Please check the configured Conjur Account.\n" \
+                      "It does not match with values from the info endpoint."
+    else
+        skip "test due to the info endpoint is not available for OSS"
+    fi
+  fi
+  assert_success
+}
+
+@test "Conjur Authenticator is valid" {
+  display_info "Validating authenticator ID from the info endpoint for enterprise"
+
+  cmd=( curl -k --connect-timeout 5 "$conjurApplianceUrl"/info -o "$TEMP_INFO_FILE")
+  display_info "Running ${cmd[@]}"
+  run "${cmd[@]}"
+  run grep name.*"$authnK8sAuthenticatorID" "$TEMP_INFO_FILE"
+
+  if [ "$status" -ne 0 ]; then
+    if (grep name "$TEMP_INFO_FILE"); then
+        display_error "Please check configured authenticator ID\n" \
+                      "It does not match with values from the info endpoint."
+    else
+        skip "test due to the info endpoint is not available for OSS"
+    fi
+  fi
+  assert_success
+}
+
 {{- end }}


helm/kubernetes-cluster-prep/templates/tests/_helm_test.bats.txt
```diff
@@ -74,4 +75,45 @@ text_color "$MAGENTA"
...
+
+  cmd=( curl -k --connect-timeout 5 "$conjurApplianceUrl"/info -o "$TEMP_INFO_FILE")
+  display_info "Running ${cmd[@]}"
+  run "${cmd[@]}"
+  run grep name.*"$authnK8sAuthenticatorID" "$TEMP_INFO_FILE"
+
+  if [ "$status" -ne 0 ]; then
+    if (grep name "$TEMP_INFO_FILE"); then
+        status=0
+        display_error "Please check configured Conjur Account.\n" \
```
diverdane, line 110: Copy/paste error, should be "...check configured authenticator ID"

helm/kubernetes-cluster-prep/templates/tests/_helm_test.bats.txt
```diff
@@ -74,4 +75,45 @@ text_color "$MAGENTA"
...
+  run grep account.*"$conjurAccount" "$TEMP_INFO_FILE"
+
+  if [ "$status" -ne 0 ]; then
+    if (grep account "$TEMP_INFO_FILE"); then
+        status=0
+        display_error "Please check the configured Conjur Account.\n" \
+                      "It does not match with values from the info endpoint"
+        assert_failure
+    else
+        skip "test due to the info endpoint is not available for OSS"
```
diverdane, line 94: Ooh, `skip` is nice.

helm/kubernetes-cluster-prep/templates/tests/_helm_test.bats.txt
```diff
@@ -74,4 +75,45 @@ text_color "$MAGENTA"
   assert_failure
 }
 {{- end }}
+
+@test "Conjur Account is valid" {
+  display_info "Validating account from the info endpoint for enterprise"
+
+  cmd=( curl -k --connect-timeout 5 "$conjurApplianceUrl"/info -o "$TEMP_INFO_FILE")
+  display_info "Running ${cmd[@]}"
+  run "${cmd[@]}"
```
diverdane, line 84: Not a big deal, but could we drop the `run` from this, i.e. just use this:
```
"${cmd[@]}"
```
since we're not using the `$status` nor `$output` from this command?

helm/kubernetes-cluster-prep/templates/tests/_helm_test.bats.txt
```diff
@@ -74,4 +75,45 @@ text_color "$MAGENTA"
...
+  run grep account.*"$conjurAccount" "$TEMP_INFO_FILE"
+
+  if [ "$status" -ne 0 ]; then
+    if (grep account "$TEMP_INFO_FILE"); then
+        status=0
```
diverdane, line 89: I think we can eliminate this `status=0` and the `assert_failure` on line 92, and just use an `assert_success` on line 96? Something like this:
```
  if [ "$status" -ne 0 ]; then
    if (grep account "$TEMP_INFO_FILE"); then
        display_error "Please check the configured Conjur Account.\n" \
                      "It does not match with values from the info endpoint"
    else
        skip "test due to the info endpoint is not available for OSS"
    fi
  fi
  assert_success
}
```

helm/kubernetes-cluster-prep/templates/tests/_helm_test.bats.txt
```diff
@@ -74,4 +75,45 @@ text_color "$MAGENTA"
...
+  display_info "Validating account from the info endpoint for enterprise"
+
+  cmd=( curl -k --connect-timeout 5 "$conjurApplianceUrl"/info -o "$TEMP_INFO_FILE")
+  display_info "Running ${cmd[@]}"
+  run "${cmd[@]}"
+  run grep name.*"$authnK8sAuthenticatorID" "$TEMP_INFO_FILE"
+
+  if [ "$status" -ne 0 ]; then
+    if (grep name "$TEMP_INFO_FILE"); then
+        status=0
```
diverdane, line 109: Same comment as above, we could drop the `status=0` and the `assert_failure`, and just use `assert_success` at the end of the test function, I think?

diverdane (commented):
This is great, I like the use of the `skip`, this is good to know.

helm/kubernetes-cluster-prep/templates/tests/_helm_test.bats.txt
rpothier: done

helm/kubernetes-cluster-prep/templates/tests/_helm_test.bats.txt
rpothier: It seems to need the `status=0` and `assert_failure`, otherwise it doesn't fail.
For the first command the status is set, but on the second command it's not getting
set automatically. I'm not sure why I didn't see it mentioned in the docs.

helm/kubernetes-cluster-prep/templates/tests/_helm_test.bats.txt
rpothier: Same reply as above

helm/kubernetes-cluster-prep/templates/tests/_helm_test.bats.txt
rpothier: updating per our offline discussion.

diverdane (approved):
LGTM!

Issue #301, opened 2021-05-17
Test app deploy helm chart also deploys postgres


izgeri:
The application deploy helm chart is missing Postgres, which the app will require to run.

The kubernetes-conjur-demo scripts use a custom build of Postgres for K8s and a RH out of the box image for OpenShift. If possible, we should try to use an OOTB image for PG in K8s too.

Note: we also want to make sure on deploy we wait for the pg container to be ready via readiness probe.

AC:
- [ ] App deploy helm chart includes postgres deploy with readiness probe for K8s apps

PR #307, opened 2021-05-18
Add reusable workflow scripts

### LINKED ISSUE #239: There are reusable scripts for development environments and automated testing
## Summary

Create scripts for development environment and CI testing:
- To generate and load authn-k8s-specific Conjur policy 
- To generate and load application-specific Conjur policy 
- To do Helm install for cluster prep and app Namespace prep 
- Use the Helm chart from #238 to deploy example application(s) with a selectable list of authenticator types to deploy 
- To verify secrets access for example application

## Example Use Cases
The scripts described here can be used for:
- Automated CI for testing Kubernetes authentication on for various Conjur OSS/Enterprise configurations
- A quick-start guide for Kubernetes authentication
- Katacoda tutorials for Kubernetes authentication

## Out of Scope
For purposes of splitting out tasks into more manageable chunks, a couple of other issues have been created to implement some aspects of this feature separately:
- __This issue does not include the addition of support for Secrets Provider init/app containers.
      Support for Secrets Provider authenticators will be added incrementally using Issue #247.__
- __This issue does not include testing the script workflow on OpenShift platform. This testing will
      be done incrementally, using Issue #248.__

## Implementation details

This issue involves basically making a copy or fork of conjurdemos/kubernetes-conjur-demo
scripts and modifying these scripts to use invocations of `helm install ...` with the following Helm charts, rather than using bash/sed/kubectl to do deployments:
   - Kubernetes cluster prep Helm chart (Issue #227)
   - Namespace prep Helm chart (Issue #236)
   - Sample application Helm chart (Issue #238)
   
The scripts for this workflow can be developed as follows:
   - Start with a clone/fork of the conjurdemos/kubernetes-conjur-demo script repository
   - The `set_env_vars.sh` script can probably be deleted. Most chart values for the
      above Helm charts will likely be required settings. (Remove its invocation from `start`).
   - Modify the `0_prep_check_dependencies.sh` file to require the existence of environment variable
      settings that correspond to every required chart value for all 3 Helm charts. For example,
      for the cluster prep Helm chart, add a check for an environment variable `CONJUR_ACCOUNT`
      corresponding to the `conjur.account` chart value. NOTE: These environment variable settings will
      be used on the `helm install ....` command line to set the corresponding chart value. For
      example, for `conjur.account`, the following command line setting will be included for
      `helm install ...`:
      
                   --set conjur.account="$CONJUR_ACCOUNT"
 
      Passing in settings to these scripts as environment variables is more convenient
      than creating a custom `values.yaml` file (and we can use Summon in CI).
        
      Note that the checks for required environment variable settings are being added here,
      even though the charts will check for required values settings themselves, so that we fail
      quickly. For example,  if we're missing env variable setting for the application deployment,
      we won't have to wait run helm install of the other Helm charts before discovering the
      missing setting.
   - Modify the `4_app_create_namespace.sh` to:
      - Eliminate the creation of the RoleBinding
      - Add invocation of `helm install ...` for cluster prep helm chart (could be a separate bash script)
      - Add invocation of `helm install ...` for Namespace prep helm chart (could be a separate bash script)
   - Delete the `5_app_store_conjur_cert.sh` script and remove its invocation from `start`
   - Modify the `7_app_deploy.sh` to use new sample Application deploy Helm chart

## DoD
- [ ] The `conjur-authn-k8s-client` project includes utility scripts for policy loading to be used in the e2e tests and in the "quick start" demo

Note: in both cases (automated tests & demo), we are deploying Conjur so that we have access to the Conjur OSS server / Conjur Enterprise leader to load policy.

imheresamir:
### What does this PR do?

This PR adds end-to-end workflow test scripts, ported over from [conjurdemos/kubernetes-conjur-demo](https://github.com/conjurdemos/kubernetes-conjur-demo) 
The ported scripts were modified to:
1) Install `conjur-oss` in a KinD cluster
2) Rework cluster config to use our new helm charts `kubernetes-cluster-config` and `app-namespace-prep`
3) Use bitnami/postgresql helm chart for the sample app backend

The `app-deploy` chart was modified to take `enabled` flags to select one or more subcharts. The `app-summon-sidecar` subchart is enabled using `--set authn-k8s.enabled=true` when installing the parent `app-deploy` chart.

The workflow can be tested by launching `bin/test-workflow/start`.

Commit 2 (947fb35) copies unchanged files from [conjurdemos/kubernetes-conjur-demo](https://github.com/conjurdemos/kubernetes-conjur-demo) (a4272e3)

Commit 3 (51a92d5) is where the script changes are.

Resolves #239 and #301

#### Change log

#### Test coverage

#### Documentation

#### Manual tests


bin/test-workflow/start
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+
+set -eo pipefail
```
diverdane, line 3: I think we'll want a README.md file to give a brief overview of what these scripts do, and how to run it (e.g. just run `./start`). This wasn't included in the DoD for the corresponding issue(s), so maybe we can file a follow-up issue for this if that works better.

There's a README.md that we might be able to grab some content:
   https://github.com/cyberark/conjur-oss-helm-chart/blob/master/examples/kubernetes-in-docker/README.md
although this README.md ^^^^ has much more verbage than we need here.

bin/test-workflow/set_env_vars.sh
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env bash
+
+# Set the default values of environment variables used by the scripts
+export PULL_DOCKER_REGISTRY_URL=${PULL_DOCKER_REGISTRY_URL:-${DOCKER_REGISTRY_URL}}
```
diverdane, line 4: At some point, we're going to be using this script in different scenarios/topologies (e.g. running in Google Cloud, or running in OpenShift, etc.). In most cases, this will probably be done by invoking the `start` script via Summon, with custom Summon YAML config for each scenario to set the environment variables that are used here.

Just a thought... feel free to skip this idea for now, but another mechanism that might come in handy for developers is to support a `customize.env` file similar to this one for the Conjur OSS KinD example:
    https://github.com/cyberark/conjur-oss-helm-chart/blob/master/examples/kubernetes-in-docker/customize.env
I've found this handy a few times when I just wanted to make small tweaks to how the example works.

If this sounds useful, this can be another followup issue, since it's low priority.

bin/test-workflow/kubernetes/conjur-cli.yml
```diff
@@ -0,0 +1,27 @@
+---
```
diverdane, line 1: Do we still need all of the YAML manifests in the `kubernetes/` and `openshift/` subdirectories? I would think all of this would be replaced with the application deploy charts/subcharts.

The one YAML manifest that we might want to keep for the short term, I suppose, is this `conjur-cli.yml` manifest, although it would be good to migrate the Conjur CLI deployment to the parent Helm chart at some point.

bin/test-workflow/7_app_deploy.sh
```diff
@@ -0,0 +1,285 @@
...
+     statefulset/summon-init-mysql \
+     statefulset/summon-sidecar-mysql \
+     statefulset/secretless-mysql \
+     secret/test-app-backend-certs
+
+  ensure_env_database
+  case "${TEST_APP_DATABASE}" in
+  postgres)
+    echo "Create secrets for test app backend"
+    $cli --namespace $TEST_APP_NAMESPACE_NAME \
```
diverdane, line 94: I think the creation of a Kubernetes Secret to store the CA cert & key can be moved to the parent application deploy Helm chart (the `ca.pem` and `ca-key.pem` files would get moved e.g. to a `files/` subdirectory in that chart.

The advantage of doing this in the parent Helm chart would be that we would have to handle this in the scripts, and the Secret would get deleted via `helm delete ...`.

diverdane (commented):
This is great stuff!!!
I'm still working my way through this, but here are the comments that I have now.

bin/test-workflow/7_app_deploy.sh
```diff
@@ -0,0 +1,285 @@
...
+
+  if is_minienv; then
+    IMAGE_PULL_POLICY='Never'
+  else
+    IMAGE_PULL_POLICY='Always'
+  fi
+
+  deploy_app_backend
+  # deploy_secretless_app
+  deploy_sidecar_app
```
diverdane, line 23: I think we need a little more logic here.

Ultimately, the choice of which of the various flavors of app integrations should be deployed should be settable by whomever is calling these scripts (i.e. it's not reusable if we're relying on commenting/uncommenting the script here). I think the best way to do this would be to add environment variables to enable each of the app integrations (and then we can control it via Summon in the Jenkins CI).

So then the question is how do we use those new environment variable settings here in the script? One way we can do this, which matches what you're doing here, me thinks, is to conditionally and separately call individual bash deploy functions based on those environment variables, and each of those bash functions would be doing separate `helm install ...` invocations. So for example, if `DEPLOY_SIDECAR_APP` env variable is `true`, then we call `deploy_sidecar_app` function, and that does a `helm install ... ` with `--set app-summon-sidecar.enabled=true` and all other integrations disabled.

This would work, but the automated testing would then be installing N different Helm releases. Not a big deal. But if we do go this route, maybe we can rethink and simplify the way we're using a main app deploy Helm chart and multiple subcharts, since we'd really be using each subchart independently (subcharts aren't buying us anything?).

Another approach would be to check each of the new environment variables, and compile a set of `--set` arguments that might add up to e.g.:
```
--set "app-summon-sidecar.enable=true" \
--set "app-summon-init.enable=false" \
--set "app-secretless-broker.enable=true" \
--set "app-secrets-provider.enable=true"
```
(along with some other app/integration specific settings), and then we would do a single `helm install ...` on the main app-deploy Helm chart.

I'm not sure what the best approach is, I'm leaning towards a single `helm install ...`, but could be persuaded otherwise.

diverdane (commented):
More comments. Still plowing through everything.

bin/test-workflow/7_app_deploy.sh
diverdane: Or another option would be to add a Kubernetes Secret for test app backends in each of the Helm subcharts, but they'd each need unique names.

bin/test-workflow/7_app_deploy.sh
```diff
@@ -0,0 +1,285 @@
...
+deploy_sidecar_app() {
+  pushd $(dirname "$0")/../../helm/app-deploy > /dev/null
+    # Deploy a given app with yet another subset of the subset of our golden configmap, allowing
+    # connection to Conjur
+    announce "Installing sidecar application chart"
+    if [ "$(helm list -q -n $TEST_APP_NAMESPACE_NAME | grep "^sidecar-app$")" = "sidecar-app" ]; then
+        helm uninstall sidecar-app -n "$TEST_APP_NAMESPACE_NAME"
+    fi
+
+    helm install sidecar-app . -n "$TEST_APP_NAMESPACE_NAME" --debug --wait \
```
diverdane, line 151: We should use `--timeout` flag for this since we're using `--wait`. The default timeout is 5 minutes. This is probably good in most cases, but it might save us at some point if we added an environment variable setting for this just in case we have some weird corner case for timing.  WDYT?

diverdane: Same for other places where we're using `--wait`.

bin/test-workflow/7_app_deploy.sh
diverdane: An advantage to using a single `helm install ...` on the parent chart is that all subcharts would be installed in parallel. Another advantage is that we'd be able to create Kubernetes resources that are shared by all of the subcharts (e.g. one common TLS Secret for all backends).

bin/test-workflow/8_app_verify_authentication.sh
```diff
@@ -0,0 +1,183 @@
...
+  # pods_ready "test-app-with-host-outside-apps-branch-summon-init" &&
+  pods_ready "test-app-summon-sidecar" # &&
+  # pods_ready "test-app-secretless"
+}
+bl_retry_constant "${RETRIES}" "${RETRY_WAIT}"  check_pods
+
+if [[ "$PLATFORM" == "openshift" ]]; then
+  echo "Waiting for deployments to become available"
+
+  check_deployment_status(){
```
diverdane, line 78: I think we can also delete the `check_deployment_status()` function since we're using `--wait` on all of the Helm installs?

bin/test-workflow/8_app_verify_authentication.sh
```diff
@@ -0,0 +1,183 @@
...
+  oc port-forward "$init_pod_with_host_outside_apps" 8084:8080 > /dev/null 2>&1 &
+  INIT_WITH_HOST_OUTSIDE_APPS_PORT_FORWARD_PID=$!
+
+  curl_cmd=curl
+  sidecar_url="localhost:8081"
+  init_url="localhost:8082"
+  secretless_url="localhost:8083"
+  init_url_with_host_outside_apps="localhost:8084"
+else
+  if [[ "$TEST_APP_LOADBALANCER_SVCS" == "true" ]]; then
```
diverdane, line 107: Maybe something to discuss this in a group meeting.... We should consider whether testing our applications using loadbalancers is giving us any real value. It seems to me that using the loadbalancer external IPs to test our applications' services versus just using a Kubernetes internal service name is testing the loadbalancer functionality of whatever Kubernetes platform we're using, not Conjur functionality.

Dropping the requirement that we need to test using external IPs would greatly simplify the testing here. For example, if we ONLY test using internal service names, then everything can be tested from inside the Kubernetes cluster (using a `curl Pod`).

This would mean that we might be able to implement the verification tests for each application integration separately, in their own Helm test. AND, it may be that running `helm test ....` on a parent chart does a Helm test on all of its subcharts????? I haven't found documentation on this.... but even if this isn't the case, we should be able to run `helm test ...` on each subchart individually.

The bottom line is that each subchart would be able to provide its own test mechanism.

bin/test-workflow/8_app_verify_authentication.sh
```diff
@@ -0,0 +1,183 @@
...
+    sidecar_url="test-app-summon-sidecar.$TEST_APP_NAMESPACE_NAME.svc.cluster.local:8080"
+    secretless_url="test-app-secretless.$TEST_APP_NAMESPACE_NAME.svc.cluster.local:8080"
+  fi
+fi
+
+echo "Waiting for urls to be ready"
+
+check_urls(){
+  (
+    # $curl_cmd -sS --connect-timeout 3 "$init_url" &&
```
diverdane, line 138: Thinking ahead to when we add more application integrations...
Ideally, these curl commands for each app integration should be run conditionally based on env variable.
However, if it's possible to move these calls into Helm tests that are defined for each subchart, then this would be a lot less fugly with conditionals scattered through this test code.

If we're not going to add conditionals (based on env variable enables) for the short term, we should at least file an issue or create a spike to either add conditional logic in, or figure out how to do this as Helm tests in each of the subcharts.

bin/test-workflow/etc/secretless.yml
```diff
@@ -0,0 +1,34 @@
+version: "2"
```
diverdane, line 1: I think this file can be deleted. It's only used for the secretless-broker integration, AND, when we add that integration, we'll probably add this as data contained in a ConfigMap, and that ConfigMap will get deployed by Helm install.

bin/test-workflow/kubernetes/mysql.template.yml
```diff
@@ -0,0 +1,137 @@
+---
```
doodlesbykumbi, line 1: I'm wondering if we're better off showing the test flow working with a single database. I don't think we gain very much by trying to capture every database. I'm flagging this because there's a lot of these manifests that we have to maintain for the different dbs.

Perhaps a better approach would be to have the app deploy help chart handle the following scenarios
1. Specify your own db credentials and app is deployed to use them
2. Specify a db type (in the helm chart values) and have the helm chart generate the database from a subchart (or manifests bundled into the helm chart). This should also be able to handle Openshift. This would make maintenance a lot simpler.

doodlesbykumbi (commented):
I think there's a lot of great improvements in the PR. I've left some comments about things that stood out, but I'm still thinking about ways in which we can eradicate the undesirable things that are inherited from the pre-existing scripts. 

I think it's good for us to really take that step back and look at this workflow from a high-level, and ask ourselves how we would implement each section given a blank slate. In this case we really do have a blank slate.

A key criteria (for myself) in considering the scripts reusable is for them to be compossible in somewhat arbitrary ways. We are constrained on how granular that decomposition by the logical boundaries of the tasks. I imagine the reusable workflow scripts should allow me to provide input to a script that can carry out any logical task listed [in the docs](ihttps://github.com/cyberark/dap-wiki/tree/master/how-to-guides/kubernetes#set-up-and-use-kubernetes-authenticator ) and provide output in a way that could potentially be piped into another script.

bin/test-workflow/7_app_deploy.sh
```diff
@@ -0,0 +1,285 @@
...
+    echo "Deploying test app backend"
+
+    # Install postgresql helm chart
+    if [ "$(helm list -q -n $TEST_APP_NAMESPACE_NAME | grep "^summon-sidecar-app-backend-pg$")" = "summon-sidecar-app-backend-pg" ]; then
+        helm uninstall summon-sidecar-app-backend-pg -n "$TEST_APP_NAMESPACE_NAME"
+    fi
+
+    kubectl delete --ignore-not-found pvc -l app.kubernetes.io/instance=summon-sidecar-app-backend-pg
+
+    helm repo add bitnami https://charts.bitnami.com/bitnami
```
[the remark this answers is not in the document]
imheresamir: After discussion with @doodlesbykumbi it was realized that it may be preferable to install the backend helm chart separately from the sample app since different personas might perform those steps

bin/test-workflow/kubernetes/conjur-cli.yml
imheresamir: Removed unused yaml files

bin/test-workflow/2_admin_load_conjur_policies.sh
```diff
@@ -0,0 +1,116 @@
...
+  announce "Deploying Conjur CLI pod."
+
+  if is_minienv; then
+    IMAGE_PULL_POLICY='Never'
+  else
+    IMAGE_PULL_POLICY='Always'
+  fi
+
+  cli_app_image=$(platform_image_for_pull conjur-cli)
+  sed -e "s#{{ CONJUR_SERVICE_ACCOUNT }}#$(conjur_service_account)#g" ./$PLATFORM/conjur-cli.yml |
```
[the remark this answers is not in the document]
imheresamir: Let's chat about this. @doodlesbykumbi doesn't want policy loading to happen within the cluster

bin/test-workflow/7_app_deploy.sh
imheresamir: Added the `enabled` flag for `app-summon-sidecar` on the parent chart

bin/test-workflow/7_app_deploy.sh
```diff
@@ -0,0 +1,285 @@
...
+        --set app-summon-sidecar.enabled=true \
+        --set global.conjur.conjurConnConfigMap="conjur-connect-configmap" \
+        --set app-summon-sidecar.conjur.authnLogin="$CONJUR_AUTHN_LOGIN_PREFIX/test-app-summon-sidecar"
+  popd > /dev/null
+
+  echo "Test app/sidecar deployed."
+}
+
+###########################
+deploy_init_container_app() {
```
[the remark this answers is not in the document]
imheresamir: Done

bin/test-workflow/7_app_deploy.sh
```diff
@@ -0,0 +1,285 @@
...
+
+  if [[ "$PLATFORM" == "openshift" ]]; then
+    oc expose service test-app-summon-init
+  fi
+
+  echo "Test app/init-container deployed."
+}
+
+###########################
+deploy_init_container_app_with_host_outside_apps() {
```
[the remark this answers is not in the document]
imheresamir: Done

bin/test-workflow/etc/secretless.yml
imheresamir: Deleted

bin/test-workflow/test_app_summon/Dockerfile.oc
```diff
@@ -0,0 +1,13 @@
+FROM cyberark/demo-app
+ARG namespace
+MAINTAINER CyberArk
+
+#---copy summon into image---#
+COPY tmp.summon-conjur /usr/local/lib/summon/summon-conjur
```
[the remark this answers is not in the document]
imheresamir: Leaving this for now since the new `demo-app` with summon included hasn't made it to DockerHub

bin/test-workflow/test_app_summon/Dockerfile.builder
```diff
@@ -0,0 +1,15 @@
+FROM ruby:2.4
```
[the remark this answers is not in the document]
imheresamir: Leaving this for now since the new `demo-app` with summon included hasn't made it to DockerHub

bin/test-workflow/stop
```diff
@@ -0,0 +1,39 @@
...
+. utils.sh
+
+KUBE_CLI_DELETE_TIMEOUT="10m"
+
+set_namespace "$TEST_APP_NAMESPACE_NAME"
+"$cli" get pods
+
+set_namespace default
+
+if [[ "$PLATFORM" == "openshift" ]]; then
```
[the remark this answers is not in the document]
imheresamir: Deleted file

bin/test-workflow/openshift/postgres.template.yml
```diff
@@ -0,0 +1,170 @@
+---
```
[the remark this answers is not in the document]
imheresamir: Deleted

bin/test-workflow/kubernetes/mysql.template.yml
imheresamir: Deleted - only testing with postgres in this round

bin/test-workflow/policy/users.yml
```diff
@@ -0,0 +1,40 @@
+---
```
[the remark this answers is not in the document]
imheresamir: Keeping this for now after our discussion

bin/test-workflow/7_app_deploy.sh
imheresamir: Added the secret creation to `app-namespace-prep` chart since it needs to be present before installing sample app backend

bin/test-workflow/test_app_summon/Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+FROM cyberark/demo-app
+ARG namespace
+MAINTAINER CyberArk
+
+#---copy summon into image---#
+COPY --from=test-app-builder /usr/local/lib/summon /usr/local/lib/summon
+COPY --from=test-app-builder /usr/local/bin/summon /usr/local/bin/summon
+
+#---copy secrets.yml into image---#
+COPY tmp.$namespace.secrets.yml /etc/secrets.yml
```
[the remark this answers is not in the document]
imheresamir: Let's discuss and create an issue

bin/test-workflow/4_kubernetes_cluster_prep.sh
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+. utils.sh
+
+set_namespace default
+
+# Prepare our cluster with conjur and authnK8s credentials in a golden configmap
+announce "Installing cluster prep chart"
+pushd $(dirname "$0")/../../helm/kubernetes-cluster-prep > /dev/null
```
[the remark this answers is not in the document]
imheresamir: Agreed. Is this ok til then?

bin/test-workflow/4_kubernetes_cluster_prep.sh
```diff
@@ -0,0 +1,22 @@
...
+    if [ "$(helm list -q -n $CONJUR_NAMESPACE | grep "^cluster-prep$")" = "cluster-prep" ]; then
+        helm uninstall cluster-prep -n "$CONJUR_NAMESPACE"
+    fi
+
+    ./bin/get-conjur-cert.sh -v -i -u "$CONJUR_APPLIANCE_URL"
+
+    helm install cluster-prep . -n "$CONJUR_NAMESPACE"  --debug --wait \
```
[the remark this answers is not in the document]
imheresamir: Done

bin/test-workflow/7_app_deploy.sh
```diff
@@ -0,0 +1,285 @@
...
+  else
+    conjur_follower_name=${CONJUR_FOLLOWER_NAME:-conjur-follower}
+    conjur_appliance_url=https://$conjur_follower_name.$CONJUR_NAMESPACE.svc.cluster.local/api
+  fi
+  conjur_authenticator_url="$conjur_appliance_url/authn-k8s/$URLENCODED_AUTHN_ID"
+}
+
+###########################
+deploy_app_backend() {
+  $cli delete --ignore-not-found \
```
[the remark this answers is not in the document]
imheresamir: Done

bin/test-workflow/test_app_summon/Dockerfile
```diff
@@ -0,0 +1,32 @@
+FROM ruby:2.4 as test-app-builder
```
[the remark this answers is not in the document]
imheresamir: Leaving this for now since the new `demo-app` with summon included hasn't made it to DockerHub

bin/test-workflow/0_prep_conjur_in_kind.sh
```diff
@@ -0,0 +1,38 @@
...
+set -eo pipefail
+
+rm -rf bash-lib
+git clone https://github.com/cyberark/bash-lib.git
+
+# Install Conjur in our cluster
+mkdir -p temp
+pushd temp > /dev/null
+    rm -rf conjur-oss-helm-chart
+    git clone https://github.com/cyberark/conjur-oss-helm-chart.git
```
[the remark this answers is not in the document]
imheresamir: Scripts were modified to decouple the conjur install. Let's chat to see if there needs to be more done here

bin/test-workflow/5_app_namespace_prep.sh
```diff
@@ -0,0 +1,20 @@
...
+
+# Prepare a given namespace with a subset of credentials from the golden configmap
+announce "Installing application namespace prep chart"
+pushd $(dirname "$0")/../../helm/application-namespace-prep > /dev/null
+    if [ "$(helm list -q -n $TEST_APP_NAMESPACE_NAME | grep "^namespace-prep$")" = "namespace-prep" ]; then
+        helm uninstall namespace-prep -n "$TEST_APP_NAMESPACE_NAME"
+    fi
+
+    # Namespace $TEST_APP_NAMESPACE_NAME will be created if it does not exist
+    helm install namespace-prep . -n "$TEST_APP_NAMESPACE_NAME"  --debug --wait \
```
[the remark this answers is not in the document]
imheresamir: Done

bin/test-workflow/6_app_build_and_push_containers.sh
```diff
@@ -0,0 +1,68 @@
...
+    if ! is_minienv; then
+      docker push $test_app_image
+    fi
+  done
+popd
+
+if [[ "$LOCAL_AUTHENTICATOR" == "true" ]]; then
+  # Re-tag the locally-built conjur-authn-k8s-client:dev image
+  authn_image=$(platform_image_for_push conjur-authn-k8s-client)
+  docker tag conjur-authn-k8s-client:dev $authn_image
```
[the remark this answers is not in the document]
imheresamir: The `conjur-authn-k8s-client` image is being used for the sidecar - let's chat about maybe leaving this for now

bin/test-workflow/8_app_verify_authentication.sh
imheresamir: Removed the loadbalancer testing option based on discussion

bin/test-workflow/utils.sh
```diff
@@ -0,0 +1,288 @@
...
+
+    while ! eval $@ > /dev/null; do
+      echo -n .
+      sleep $spacer
+    done
+    echo 'Success!'
+  fi
+}
+
+function is_minienv() {
```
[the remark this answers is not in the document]
imheresamir: Done

bin/test-workflow/start
```diff
@@ -0,0 +1,18 @@
...
+
+cd "$(dirname "$0")" || ( echo "cannot cd into dir" && exit 1 )
+
+. ./0_prep_conjur_in_kind.sh
+
+. ./1_prep_check_dependencies.sh
```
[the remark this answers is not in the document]
imheresamir: Was needed before but changed now

bin/test-workflow/8_app_verify_authentication.sh
imheresamir: Done

bin/test-workflow/8_app_verify_authentication.sh
```diff
@@ -0,0 +1,183 @@
...
+if [[ "$TEST_APP_LOADBALANCER_SVCS" == "false" ]]; then
+  echo "Deploying a test curl pod"
+  deploy_test_curl
+  echo "Waiting for test curl pod to become available"
+  bl_retry_constant "${RETRIES}" "${RETRY_WAIT}"  check_test_curl
+fi
+  
+echo "Waiting for pods to become available"
+
+check_pods(){
```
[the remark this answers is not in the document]
imheresamir: Done

bin/test-workflow/7_app_deploy.sh
imheresamir: Done

bin/test-workflow/start
imheresamir: Created #314

bin/test-workflow/7_app_deploy.sh
imheresamir: Reverted creation of Secret in a helm chart after discussion with @diverdane

bin/test-workflow/6_app_build_and_push_containers.sh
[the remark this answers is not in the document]
imheresamir: Cleaned up

bin/test-workflow/7_app_deploy.sh
```diff
@@ -0,0 +1,285 @@
...
+        --set volumePermissions.enabled=true \
+        --set tls.certificatesSecret="test-app-backend-certs" \
+        --set tls.certFilename="server.crt" \
+        --set tls.certKeyFilename="server.key" \
+        --set securityContext.fsGroup="999" \
+        --set postgresqlDatabase="test_app" \
+        --set postgresqlUsername="test_app" \
+        --set postgresqlPassword=$SAMPLE_APP_BACKEND_DB_PASSWORD
+    ;;
+  mysql)
```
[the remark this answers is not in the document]
imheresamir: Added #315

bin/test-workflow/8_app_verify_authentication.sh
imheresamir: Added #316

bin/test-workflow/set_env_vars.sh
```diff
@@ -0,0 +1,29 @@
...
+LOCAL_AUTHENTICATOR="${LOCAL_AUTHENTICATOR:-false}"
+
+# Some older workflows that use this script repo may depend upon
+# the the use of 'DEPLOY_MASTER_CLUSTER' environment variable rather than
+# the newer (and more accurately named) 'CONFIGURE_CONJUR_MASTER'.
+DEPLOY_MASTER_CLUSTER="${DEPLOY_MASTER_CLUSTER:-false}"
+CONFIGURE_CONJUR_MASTER="${CONFIGURE_CONJUR_MASTER:-$DEPLOY_MASTER_CLUSTER}"
+
+ANNOTATION_BASED_AUTHN="${ANNOTATION_BASED_AUTHN:-true}"
+CONJUR_OSS_HELM_INSTALLED="${CONJUR_OSS_HELM_INSTALLED:-true}"
```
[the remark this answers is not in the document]
imheresamir: Added #317

bin/test-workflow/test_app_summon/Dockerfile.oc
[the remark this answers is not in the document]
imheresamir: Created #318

doodlesbykumbi (approved):
Approved with some follow up issues in mind, here's a rough list

1. Rip out platform branching
2. Revisit platform branching. Helm is great for this, `openshift=true`  can be used to toggle between openshift and k8s in the spirit of the bitnami postgres chart
3. Update policy loading to match dap-wiki (simpler)
4. Refactor scripts in a way that allows personas to execute steps independent of each other in a way that aligns with dap-wiki
5. Refactor wholesale loading of policy, to allow for loading of policy and writing of variables local to scripts and using the identity of the relevant persona

PR #330, opened 2021-06-11
This adds a Helm release validation tests for the Helm charts.

### LINKED ISSUE #241: There are helm release validation tests for the Conjur cluster prep helm charts
Helm release validation tests will make use of the [Helm Test feature](https://helm.sh/docs/helm/helm_test/) to validate deployed instances (releases) of Helm charts. 

Helm release validation tests are available for both the Kubernetes cluster prep Helm chart and the application Namespace prep Helm chart. 

The list of automated test cases below include both positive test cases (for which Helm test success is expected) and negative test cases (for which Helm test failure is expected, e.g. for incorrect configuration). In essence, we are “validating the validator” for our Helm release validation tests. 

The overall workflow for this automated testing is: 
- Create a Kubernetes cluster for testing. (This can be done using Kubernetes-in-Docker, a.k.a. KinD, in a GitHub action). 
- Create a Conjur instance. (This can be done using Conjur OSS Helm chart). 
- Generate and load Conjur policy for the authn-k8s authenticator. 
- Run `git clone …` to get a local copy of the Kubernetes cluster prep Helm chart. 
- Generate and load application-specific Conjur policy. 
- Run `git clone …` to get a local copy of the application Namespace prep Helm chart. 
- For each test case listed below: 
   - Create an appropriate, custom `values.yaml` file for the Kubernetes cluster prep Helm chart based on the test case scenario. 
   - Run `helm install …` for the Kubernetes cluster prep Helm chart using the modified `values.yaml`. 
   - Run `helm test` for the target Helm chart (second column in the chart below). 
   - Verify `helm test` results are as expected for this test case. 
   - Run `helm delete …` for Kubernetes cluster prep Helm chart. 

## Kubernetes cluster prep helm chart test scenarios
- [ ] Happy path: With correct chart values, Helm test passes (test Pod can authenticate with Conjur) 
- [ ] Incorrect Conjur URL: Helm test fails with indication that it cannot curl Conjur 
- [ ] Incorrect Conjur SSL certificate: Helm test fails with indication that configured Conjur SSL cert does not match Conjur’s actual SSL cert. 
- [ ] Incorrect Authenticator ID: Helm test fails with indication that the test Pod cannot authenticate with Conjur

rpothier:
### What does this PR do?
This adds a Helm release validation tests.
This script will create a Kind cluster, run the Helm cluster and namespace
prep Helm charts and run Helm tests.
There is also negative tests to validate the Helm tests fail as expected.

Resolves #241

#### Change log

#### Test coverage

#### Documentation

#### Manual tests

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,0 +1,56 @@
+name: CI
+
+on:
+  # Run this on pushes to main
+  push:
+    branches:
+    - main
+
+  # Or when PR operations are done
+  pull_request:
+    types:
+    - opened
+    - reopened
+    - synchronize
+
+jobs:
+  validate-test-helm-v3:
+    name: Validate Helm tests with Helm V3 on a KinD Cluster
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        kube-tag:
+          - v1.18.2
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v2
+
+      - name: Install Helm
+        uses: azure/setup-helm@v1
+        with:
+          version: v3.3.0
+
+      - name: Run Helm happy test
+        run: ./bin/validate-helm -c -t happy > helm.log
+
+      - name: Run Helm incorrect_url test
+        run: ./bin/validate-helm -t incorrect_url > helm.log
+
+      - name: Check status
+        if: ${{ always() }}
+        run: tail helm.log
+
+      - name: Run Helm incorrect_cert test
+        run: ./bin/validate-helm -t incorrect_cert > helm.log
+
+      - name: Check status
+        if: ${{ always() }}
+        run: tail helm.log
+
+      - name: Run Helm incorrect_auth_id test
+        run: ./bin/validate-helm -t incorrect_auth_id > helm.log
+
+      - name: Check status
+        if: ${{ always() }}
+        run: tail helm.log
--- a/bin/validate-helm
+++ b/bin/validate-helm
@@ -1,0 +1,181 @@
+#!/bin/bash
+
+cd "$(dirname "$0")"
+
+source ./test-workflow/utils.sh
+source ../helm/common/utils.sh
+
+helm_tests_return_code=0
+
+tests=(happy incorrect_url incorrect_cert incorrect_auth_id)
+
+print_usage() {
+  echo "Usage:"
+  echo "    This script will validate the helm chart tests"
+  echo ""
+  echo "Syntax:"
+  echo "    $0  [Options]"
+  echo "    Options:"
+  echo "    -h            Show help"
+  echo "    -c            Create the Kind cluster"
+  echo "    -d            Delete the Kind cluster"
+  echo "    -t            run one test"
+}
+
+
+function pushd() {
+    command pushd "$@" > /dev/null
+}
+
+function popd() {
+    command popd "$@" > /dev/null
+}
+
+function create_cluster() {
+# Create kind cluster
+    ./0_prep_conjur_in_kind.sh
+    . ./1_prep_env.sh
+    ./2_admin_load_conjur_policies.sh
+    ./3_admin_init_conjur_cert_authority.sh
+}
+
+function delete_cluster() {
+
+    kind delete cluster
+}
+
+readonly APPLIANCE_URL="https://conjur-oss.conjur-oss.svc.cluster.local"
+
+function clone_ws_and_get_cert() {
+    rm -rf conjur-authn-k8s-client
+    git clone https://github.com/cyberark/conjur-authn-k8s-client.git
+    pushd conjur-authn-k8s-client/helm/conjur-config-cluster-prep
+    ./bin/get-conjur-cert.sh -u $APPLIANCE_URL -i -v -s
+    popd
+}
+
+function create_cluster_helm_chart() {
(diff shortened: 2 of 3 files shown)


.github/workflows/ci.yaml
```diff
@@ -0,0 +1,61 @@
...
+
+      - name: Install Helm
+        uses: azure/setup-helm@v1
+        with:
+          version: v3.3.0
+
+      - name: Install Helm
+        uses: azure/setup-helm@v1
+        with:
+          version: v3.2.1
```
diverdane, line 37: I think we can skip install of Helm 3.2.1, since 3.3.0 is also installed?

helm/conjur-config-cluster-prep/templates/tests/_helm_test.bats.txt
```diff
@@ -50,7 +50,7 @@ function conjur_access_token_exists() {
   run conjur_is_reachable
   if [ "$status" -ne 0 ]; then
     display_error "The 'conjur.applianceURL' chart value is set to\n" \
```
diverdane, line 52: Thanks for cleaning up line 53. If you're making any other changes, the `applianceURL` should also be changed to `applianceUrl` on line 52.

bin/validate-helm
```diff
@@ -0,0 +1,185 @@
...
+}
+
+function delete_cluster() {
+
+    kind delete cluster
+}
+
+readonly APPLIANCE_URL="https://conjur-oss.conjur-oss.svc.cluster.local"
+
+function git_clone() {
```
diverdane, line 49: Looks like the `git_clone()` function clones the authn-k8s repo AND gets the Conjur cert. Should this function be renamed to reflect that its main purpose is to get the Conjur cert?

bin/validate-helm
```diff
@@ -0,0 +1,185 @@
...
+    rm -rf conjur-authn-k8s-client
+    git clone https://github.com/cyberark/conjur-authn-k8s-client.git
+    pushd conjur-authn-k8s-client/helm/conjur-config-cluster-prep
+    ./bin/get-conjur-cert.sh -u $APPLIANCE_URL -i -v -s
+}
```
diverdane, line 54: Should there be a `popd` here in `git_clone()`? i.e. have the function restore state as much as it can when it has done its thing?

bin/validate-helm
```diff
@@ -0,0 +1,185 @@
...
+
+  pushd test-workflow
+
+  echo $(pwd)
+
+  if [ "$create_kind" = true ] ; then
+    create_cluster
+  fi
+  mkdir -p temp_verify
+  pushd temp_verify > /dev/null
```
diverdane, line 153: The redirect to `/dev/null` can be removed, if `pushd` is calling your helper function that's doing the same.

bin/validate-helm
```diff
@@ -0,0 +1,185 @@
...
+       banner "$RED" "Helm install Failed"
+  fi
+  helm get values "$helm_release"
+}
+
+function test_helm_chart() {
+
+  release=$1
+  test_name=$2
+  expected=1
```
diverdane, line 101: Can we be certain that `helm test ...` will always return a 1 upon failure, or just that it will be non-zero?
Might take a few more conditionals, but would be safer to check for zero vs. non-zero?

diverdane (commented):
Looking good!!! A few minor suggestions/comments.

helm/conjur-config-cluster-prep/templates/tests/_helm_test.bats.txt
```diff
@@ -50,7 +50,7 @@ function conjur_access_token_exists() {
...
-                  "$conjurApplianceURL. This is not reachable via 'curl -k'"
+                  "$conjurApplianceUrl. This is not reachable via 'curl -k'"
```
doodlesbykumbi, line 53: I think we should consider updating the messaging to something that gives some explanation for the reachability  
```
This is not reachable outside the Kubernetes network, so running 'curl -k ...' on your local machine probably won't work!
```

helm/conjur-config-cluster-prep/templates/tests/_helm_test.bats.txt
rpothier: Changed line 52 also.

helm/conjur-config-cluster-prep/templates/tests/_helm_test.bats.txt
rpothier: It could be internal to the K8s network, and just a config error. That's what we are looking to detect.

.github/workflows/ci.yaml
rpothier: yes, we don't need to downgrade.

bin/validate-helm
rpothier: good idea.

helm/conjur-config-cluster-prep/templates/tests/_helm_test.bats.txt
diverdane: To add to that... the `curl -k <Conjur-appliance-URL>` is always executed from a Pod inside the K8s cluster because this is a Helm test, but the destination Conjur appliance URL can be an internal K8s address or an external-to-k8s address.

diverdane (approved):

PR #334, opened 2021-06-16
Bump version to 0.20.0

micahlee:
This PR bumps the version to `0.20.0` for release.

### DIFF
--- a/pkg/authenticator/version.go
+++ b/pkg/authenticator/version.go
@@ -4,7 +4,7 @@ import "fmt"
 
 // Version field is a SemVer that should indicate the baked-in version
 // of the authn-k8s-client
-var Version = "0.19.1"
+var Version = "0.20.0"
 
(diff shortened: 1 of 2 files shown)


pkg/authenticator/version.go
```diff
@@ -4,7 +4,7 @@ import "fmt"
 
 // Version field is a SemVer that should indicate the baked-in version
 // of the authn-k8s-client
-var Version = "0.19.1"
+var Version = "0.19.2"
```
orenbm, line 7: i think that the `RH base image` change (which does not have a linked issue btw) is a change that requires a MINOR version bump so we should change the version to `0.20.0`).

wdyt?

micahlee: Works for me. I also added the missing PR link.

orenbm (approved):
LGTM! thanks for also adding the link to the change!

PR #342, opened 2021-06-25
Change namespace-prep chart version to 0.1.0

imheresamir:
Change namespace-prep chart version to 0.1.0 to match cluster-prep


diverdane (approved):
Thanks for fixing this!!!

PR #350, opened 2021-07-07
Add GitHub Action which generates manifests from Helm charts

### LINKED ISSUE #351: Generate K8S manifests from Helm charts
Raw Kubernetes manifests should be generated from the `conjur-config-cluster-prep` and `conjur-config-namespace-prep` Helm charts so that personas configuring a Kubernetes cluster can execute the same simplified workflow without needing Helm (instead using plain `kubectl apply` commands).

Ideally these raw manifests would be generated directly from the Helm charts using `helm template`. Consider adding a Github Action to generate these manifests automatically whenever the Helm charts change.

AC:
- [ ] Manifests should be checked into a public GitHub repo that can be referenced in official docs.
- [ ] Manifests should also be bundled into a single archive (e.g. zip) for each Helm release (e.g. GitHub releases could work).

imheresamir:
### What does this PR do?

Two new Github Actions added:
1) Auto-commit raw k8s manifests generated via `helm template` from `cluster-prep` and `namespace-prep` Helm charts
2) Upload raw manifests to tagged releases

Resolves #351 and #352

#### Change log

#### Test coverage

#### Documentation

#### Manual tests

### DIFF
--- a/.github/workflows/helm-template-generate-k8s-manifests.yaml
+++ b/.github/workflows/helm-template-generate-k8s-manifests.yaml
@@ -1,0 +1,61 @@
+name: Generate K8S Manifests from Helm Charts
+
+on:
+  # Run this when PR operations are done
+  pull_request:
+    types: [ opened, synchronize, reopened, edited ]
+
+jobs:
+  generate-k8s-manifests:
+    name: Generate K8S manifests via `helm template` command
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        kube-tag:
+          - v1.21.3
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v2
+
+      - name: Configure Git
+        run: |
+          git config --local user.email "action@github.com"
+          git config --local user.name "GitHub Action"
+      
+      - name: Install kubectl
+        uses: azure/setup-kubectl@v1
+        with:
+          version: v1.21.3
+        id: install
+
+      - name: Install Helm
+        uses: azure/setup-helm@v1
+        with:
+          version: v3.6.2
+      
+      - name: Install Homebrew
+        id: set-up-homebrew
+        uses: Homebrew/actions/setup-homebrew@master
+
+      - name: Install yq
+        run: brew install yq
+
+      - name: Generate manifest for cluster-prep
+        run: bin/helm-generate-raw-manifest-cluster-prep.sh
+
+      - name: Generate manifest for namespace-prep
+        run: bin/helm-generate-raw-manifest-namespace-prep.sh
+
+      - name: Commit generated manifests
+        id: commit-generated-manifests
+        run: |
+          git add helm/conjur-config-cluster-prep/generated
+          git add helm/conjur-config-namespace-prep/generated
+          git commit -m "Auto-commit of K8S manifests generated from Helm charts"
+        continue-on-error: true
+
+      - name: Push files
+        if: steps.commit-generated-manifests.outcome == 'success'
+        run: |
+          git push --force "https://${{ github.actor }}:${{secrets.GITHUB_TOKEN}}@github.com/${{ github.repository }}.git" "HEAD:$GITHUB_HEAD_REF"
--- a/bin/helm-generate-raw-manifest-cluster-prep.sh
+++ b/bin/helm-generate-raw-manifest-cluster-prep.sh
@@ -1,0 +1,19 @@
+#!/bin/bash
+
+set -eo pipefail
+
+cd "$(git rev-parse --show-toplevel)"
+
+pushd helm/conjur-config-cluster-prep > /dev/null
+    mkdir -p generated
+
+    helm template cluster-prep . \
+        -f sample-values.yaml \
+        --render-subchart-notes \
+        --skip-tests > generated/conjur-config-cluster-prep.yaml
+
+    yq eval -i \
+        'del(.data.conjurSslCertificateBase64)' \
+        generated/conjur-config-cluster-prep.yaml
+
+popd > /dev/null
--- a/helm/conjur-config-cluster-prep/generated/conjur-config-cluster-prep.yaml
+++ b/helm/conjur-config-cluster-prep/generated/conjur-config-cluster-prep.yaml
@@ -1,0 +1,74 @@
+---
+# Source: conjur-config-cluster-prep/templates/serviceaccount.yaml
+# This ServiceAccount provides the Kubernetes RBAC identity for the Conjur Kubernetes authenticator
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: conjur-serviceaccount
+  labels:
+    release: cluster-prep
+    heritage: Helm
+    app.kubernetes.io/name: "conjur-serviceaccount"
+    app.kubernetes.io/component: "conjur-kubernetes-identity"
+    app.kubernetes.io/instance: "conjur-serviceaccount"
+    app.kubernetes.io/part-of: "conjur-config"
+    conjur.org/name: "conjur-serviceaccount"
+    helm.sh/chart: conjur-config-cluster-prep-0.1.0
+---
+# Source: conjur-config-cluster-prep/templates/golden_configmap.yaml
+# The Golden ConfigMap keeps a reference copy of Conjur configuration information
+# that will be used for subsequent operations such as preparing application Namespaces
+# for using Conjur Kubernetes authentication.
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: conjur-configmap
+  labels:
+    release: cluster-prep
+    heritage: Helm
+    app.kubernetes.io/name: "conjur-golden-configmap"
+    app.kubernetes.io/component: "conjur-reference-config"
+    app.kubernetes.io/instance: "conjur-golden-configmap"
+    app.kubernetes.io/part-of: "conjur-config"
+    conjur.org/name: "conjur-golden-configmap"
+    helm.sh/chart: conjur-config-cluster-prep-0.1.0
+data:
+  # authn-k8s Configuration 
+  authnK8sAuthenticatorID: <Insert-Authenticator-ID-Here>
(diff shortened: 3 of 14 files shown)


.github/workflows/helm-template-generate-k8s-manifests.yaml
```diff
@@ -0,0 +1,64 @@
+name: Generate K8S Manifests from Helm Charts
+
+on:
+  # Run this when PR operations are done
+  pull_request:
+    types: [ opened, synchronize, reopened, edited ]
```
diverdane, line 6: Is there any chance that this action can cause recursion (e.g. PR gets pushed, an auth-push happens for generated files, causing another GH action... )? Or is the auto-push operation essentially a no-op if there are no changes to the generated manifests?

.github/workflows/helm-template-generate-k8s-manifests.yaml
```diff
@@ -0,0 +1,64 @@
...
+          git config --local user.name "GitHub Action"
+
+      - name: Install Helm
+        uses: azure/setup-helm@v1
+        with:
+          version: v3.6.2
+
+      - name: Generate manifest for cluster-prep
+        run: |
+          pushd helm/conjur-config-cluster-prep > /dev/null
```
diverdane, line 33: Just a thought... it might be good to have the manifest generation script available locally, e.g. in the Helm chart root directory?

.github/workflows/helm-template-generate-k8s-manifests.yaml
```diff
@@ -0,0 +1,64 @@
...
+        run: |
+          git add helm/conjur-config-cluster-prep/generated
+          git add helm/conjur-config-namespace-prep/generated
+          git commit -m "Auto-commit of K8S manifests generated from Helm charts"
+        continue-on-error: true
+
+      - name: Push files
+        if: steps.commit-generated-manifests.outcome == 'success'
+        run: |
+          git push --force "https://${{ github.actor }}:${{secrets.GITHUB_TOKEN}}@github.com/${{ github.repository }}.git" "HEAD:$GITHUB_HEAD_REF"
```
doodlesbykumbi, line 64: The `--force` seems dangerous! Is it necessary ?

doodlesbykumbi (changes requested):

.github/workflows/helm-template-generate-k8s-manifests.yaml
```diff
@@ -0,0 +1,64 @@
...
+        with:
+          version: v3.6.2
+
+      - name: Generate manifest for cluster-prep
+        run: |
+          pushd helm/conjur-config-cluster-prep > /dev/null
+            mkdir -p generated
+
+            helm template cluster-prep . \
+              -f user-values.yaml \
```
diverdane, line 37: Would it be better to use a name other than `user-values.yaml` that indicates how it's used, maybe `sample-values.yaml` or something like that?

helm/conjur-config-cluster-prep/generated/conjur-config-cluster-prep.yaml
```diff
@@ -0,0 +1,258 @@
...
+    authnK8sNamespace: default
+    authnK8sServiceAccount: conjur-serviceaccount
+
+    # Conjur Configuration 
+    conjurAccount: myConjurAccount
+    conjurApplianceUrl: https://conjur-oss.conjur-oss.svc.cluster.local
+    conjurSslCertificate: ""
+    conjurSslCertificateBase64: ""
+---
+# Source: conjur-config-cluster-prep/templates/tests/test-cluster-prep-configmap.yaml
```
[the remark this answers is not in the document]
diverdane: I believe this can be addressed by adding the test hook annotations to the test ConfigMap and RoleBinding, similar to what we have for the test Pod:

```
  annotations:
    "helm.sh/hook": test-success
```

.github/workflows/helm-template-generate-k8s-manifests.yaml
```diff
@@ -0,0 +1,64 @@
...
+              --skip-tests | tee generated/conjur-config-namespace-prep.yaml
+          popd > /dev/null
+
+      - name: Commit generated manifests
+        id: commit-generated-manifests
+        run: |
+          git add helm/conjur-config-cluster-prep/generated
+          git add helm/conjur-config-namespace-prep/generated
+          git commit -m "Auto-commit of K8S manifests generated from Helm charts"
+        continue-on-error: true
```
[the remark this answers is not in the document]
diverdane: Scratch my comment above... If I understand correctly, the `continue-on-error: true` is needed so that the Jenkins job is not going to fail when the `git commit -m ...` fails, since this is going to be the normal, typical case where there are no changes to the generated manifests.

.github/workflows/helm-template-generate-k8s-manifests.yaml
```diff
@@ -0,0 +1,64 @@
...
+              --render-subchart-notes \
+              --skip-tests | tee generated/conjur-config-namespace-prep.yaml
+          popd > /dev/null
+
+      - name: Commit generated manifests
+        id: commit-generated-manifests
+        run: |
+          git add helm/conjur-config-cluster-prep/generated
+          git add helm/conjur-config-namespace-prep/generated
+          git commit -m "Auto-commit of K8S manifests generated from Helm charts"
```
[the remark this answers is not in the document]
imheresamir: `git commit` returns a nonzero exit code if there are no changes to the generated files - the `git push` in that case would become a no-op

.github/workflows/helm-template-generate-k8s-manifests.yaml
imheresamir: > the auto-push operation essentially a no-op if there are no changes to the generated manifests?
^This

.github/workflows/helm-template-generate-k8s-manifests.yaml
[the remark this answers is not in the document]
imheresamir: Yep exactly!

helm/conjur-config-cluster-prep/generated/conjur-config-cluster-prep.yaml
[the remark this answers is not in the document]
imheresamir: Done - used `success` instead of `test-success` based on latest Helm v3 guidance

.github/workflows/helm-template-generate-k8s-manifests.yaml
imheresamir: Done

.github/workflows/helm-template-generate-k8s-manifests.yaml
imheresamir: Done

.github/workflows/helm-template-generate-k8s-manifests.yaml
imheresamir: It does seem to be necessary. Shouldn't be dangerous since we're only committing generated files

helm/conjur-config-cluster-prep/sample-values.yaml
```diff
@@ -0,0 +1,11 @@
+# User-supplied values for the conjur-config-cluster-prep Helm chart.
+# This is a YAML-formatted file.
+# Declare variables to be passed into your templates.
+
+conjur:
+  account: "<Insert Conjur Account Here>"
+  applianceUrl: "https://insert.conjur.appliance.url.here"
```
diverdane, line 7: If we're going to be doing any post-processing of the generated, raw manifests, it would be helpful to wrap the `applianceUrl` and `certificateBase64` canned values in angled brackets.

Otherwise, I'm good with the canned values used here if @alexkalish is.

diverdane (dismissed):
LGTM!!!
It would be nice if the "canned" values for `applianceUrl` and `certificateBase64` were encapsulated in angled brackets to mark them clearly as "fill-in-the-blanks", but I'm okay with the output @alexkalish is!

helm/conjur-config-cluster-prep/sample-values.yaml
imheresamir: Angled brackets for all canned values would be ideal, but the schema we set for the url field for example will complain if the value isn't valid. Maybe there's a way to skip the schema validation?

diverdane: I don't know of an easy way to skip the schema validation... however, this sequence might work:

- Run helm template to generate raw manifests
- Run kustomize to fix up the "fill-in-the-blank" values in the generated raw manifests.

imheresamir: Used `yq` instead of `kustomize` since the latter does not preserve comments

bin/helm-generate-raw-manifest-cluster-prep.sh
```diff
@@ -0,0 +1,19 @@
...
+
+pushd helm/conjur-config-cluster-prep > /dev/null
+    mkdir -p generated
+
+    helm template cluster-prep . \
+        -f sample-values.yaml \
+        --render-subchart-notes \
+        --skip-tests > generated/conjur-config-cluster-prep.yaml
+
+    yq eval -i \
```
diverdane, line 15: Cool!!! This is much better than archaic sed/awk.

diverdane: Maybe as a followup to this (certainly not for this PR), maybe we can use `yq` to add angled brackets around the "fill-in-your-values" fields in the generated templates?

Or is that more trouble than it's worth?

diverdane (approved):
LGTM!!!!  Very nice work getting all of the parts working!!!

bin/helm-generate-raw-manifest-cluster-prep.sh
imheresamir: Hmm yeah that would work - although then wouldn't it be a bit confusing that we have 2 sources of sample values: 1 for rendering via helm template (`sample-values.yaml`) and another from somewhere else just to put the values in angled brackets?

PR #363, opened 2021-08-02
Yaml comments for K8s resources 

### LINKED ISSUE #353: Add YAML comments describing K8S resources created by Helm charts
> Manifests have detailed comments that help readers better understand their purpose and content. Minimally, this is a overview comment at the top of each resource, but ideally, includes some inline comments as well.

Add a descriptive YAML comment to each K8S resource created by the `cluster-prep` and `namespace-prep` Helm charts such that they appear in the autogenerated raw manifests.

NaomiHaser:
### What does this PR do?
Add a descriptive YAML comment to each K8S resource created by the cluster-prep and namespace-prep Helm charts such that they appear in the autogenerated raw manifests.

Resolves #353

#### Change log

#### Test coverage

#### Documentation

#### Manual tests

### DIFF
--- a/helm/conjur-config-cluster-prep/templates/clusterrole.yaml
+++ b/helm/conjur-config-cluster-prep/templates/clusterrole.yaml
@@ -1,3 +1,4 @@
+# This ClusterRole defines the Kubernetes API access permissions that the Conjur authenticator will require in order to validate application identities.
 {{- if .Values.authnK8s.clusterRole.create }}
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
--- a/helm/conjur-config-cluster-prep/templates/clusterrole.yaml
+++ b/helm/conjur-config-cluster-prep/templates/clusterrole.yaml
@@ -1,5 +1,6 @@
-# This ClusterRole defines the Kubernetes API access permissions that the Conjur authenticator will require in order to validate application identities.
 {{- if .Values.authnK8s.clusterRole.create }}
+# This ClusterRole defines the Kubernetes API access permissions that the Conjur
+# authenticator will require in order to validate application identities.
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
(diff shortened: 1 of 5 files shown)


helm/conjur-config-cluster-prep/templates/clusterrole.yaml
```diff
@@ -1,3 +1,4 @@
+# This ClusterRole defines the Kubernetes API access permissions that the Conjur authenticator will require in order to validate application identities.
```
diverdane, line 1: This comment should be moved below the next line (i.e. it should be inside the `{{- if .Values.authnK8s.clusterRole.create }}` conditional). Having this comment outside of the conditional is causing the CI tests to break for any test cases where the conditional is false. When the conditional is false, Helm will still render the comment but nothing else... and when it goes to install this standalone comment, it errors out because there's no API version provided in this comment-only manifest.

diverdane, line 1: Would you mind splitting up all of the comments into 80-character (or thereabouts) line lengths (preceding each line in the block comment with a `#`)? The line length you have is valid YAML, but in all of our other code types, we generally keep max line length to about 80 chars.

helm/conjur-config-cluster-prep/templates/golden_configmap.yaml
```diff
@@ -1,3 +1,5 @@
+# The Golden ConfigMap keeps a reference copy of Conjur configuration information that will be used for subsequent operations such as preparing application Namespaces for using Conjur Kubernetes authentication.
+
```
diverdane, line 2: Probably neater to not include empty lines after your comments, since these empty lines will appear in the rendered manifests.

diverdane (changes requested):
Wording looks good! Comments need to be moved into conditional blocks in order to get CI to pass.

helm/conjur-config-cluster-prep/templates/clusterrole.yaml
NaomiHaser: Ohh makes sense!

diverdane (approved):
LGTM!!!

PR #368, opened 2021-08-06
Add hook weights to helm test resources where needed

imheresamir:
### What does this PR do?

- Raw manifest generator GitHub Action removed and to be superseded by #370
- The Helm test hooks are reintroduced to the chart test resources
- Hook weights are added to non-Pod resources such that these resources get created before the Pods do as a workaround for [this bug](https://github.com/helm/helm/issues/10006) in Helm which messes up the resource creation order
- Helm test scripts cleaned up a bit to remove usage of `helm test --logs` which has [a bug](https://github.com/helm/helm/issues/8342)

#### Change log

#### Test coverage

#### Documentation

#### Manual tests


helm/conjur-config-cluster-prep/templates/tests/test-cluster-prep-configmap.yaml
```diff
@@ -2,6 +2,9 @@ apiVersion: v1
 kind: ConfigMap
 metadata:
   name: {{ .Release.Name }}-tests-configmap
+  annotations:
+    "helm.sh/hook": test
+    "helm.sh/hook-weight": "-1"
```
diverdane, line 7: Can you add a helpful reference to the Helm bug that explains why the `-1` hook weight is required into the PR description?

diverdane (commented):
Looks good!!!
One comment, and a couple of questions:

- Should the template generation GH action script be deleted?
- Do we need a new GH action to check whether template generation needs to be done?

helm/conjur-config-cluster-prep/templates/tests/test-cluster-prep-configmap.yaml
imheresamir: Done

diverdane (approved):
LGTM!!!
Thanks for filing the followup issue, I take it we'll need to add a docker run of the generate command? I think there's already a test image to do that, although it's probably missing `yq`.

imheresamir:
>  I take it we'll need to add a docker run of the generate command? I think there's already a test image to do that, although it's probably missing yq.
Yep! I'll add that to the new issue description :)

Issue #346, opened 2021-06-28
Application deployment on OpenShift


jtuttle:
The app deploy scripts will also need some OpenShift-specific modifications
for app-specific Conjur policies that are loaded, and for deploying applications.
This includes:

- In the policy subdirectory, add Conjur policy host definitions for sample applications that will be run on OpenShift.
- Using the conjurdemos/kubernetes-conjur-demo scripts as a reference, make sure that the calls to the `oc` CLI that appear in the following scripts are included in the scripts created by Issue #239:
  - 1_prep_platform_login.sh     (`oc login ...` into platform)
  - 4_app_create_namespace.sh  (`oc adm policy ...` and `oc login ...`)
  - 7_app_deploy.sh     (`oc delete ...` and `oc expose ...`)
  - 8_app_verify_authentication.sh   (`oc port-forward ...`)

## DoD
- [ ] app deploy scripts work on OpenShift platform

Issue #272, opened 2021-04-06
The app deploy Helm chart supports deployment of app + Secrets Provider init container


diverdane:
## Is your feature request related to a problem? Please describe.

Issue #238 creates a Helm chart and Helm subchart for deploying an application/Summon + authn-k8s sidecar container.

This issue expands on the application deployment main Helm chart by adding a subchart to deploy an application + Secrets Provider as an init container.

Refer to Issue #238 for implementation details for how a subchart can be added to the main app deployment Helm chart.

## Describe the solution you would like

A Helm subchart is added to the app deployment Helm chart for deploying an application + Secrets Provider sidecar.

## Out of Scope
- Creating subchart for app + Secretless Broker sidecar container
- Creating subchart for app + Secrets Provider standalone
- Creating a `README.md` file for the main Helm chart or the subchart
- Creating a `values.schema.json` for the main Helm chart or the subchart.
- Development of a Helm test for this chart is not included in this issue.
- Development of a helm-unittest for this chart is not included in this issue.

## DoD
- [ ] Helm `values.yaml` created for app subchart
- [ ] Helm manifest templates created for application + Secrets Provider sidecar, including secrets mapping ConfigMap
- [ ] Helm chart successfully deploys (application + Secrets Provider sidecar container).
   This Helm deployment assumes that the following is already available (not part of this issue):
   - Conjur instance with application policy already loaded
   - Kubernetes cluster prep Helm chart has already been Helm installed
   - Application Namespace prep Helm chart has already been Helm installed
- [ ] Helm chart includes a `templates/NOTES.txt` that announces app/authenticator has been deployed

PR #378, opened 2021-08-13
Include Secrets Provider init app in E2E workflow

john-odonnell:
### What does this PR do?
In #272, the app deploy Helm Chart was updated to support deployment of the app + Secrets Provider init container.
Here, the E2E workflow scripts are updated to deploy and test against this application.
Also, updates the app + Secrets Provider init container subchart for Openshift deployment, similar to changes to the `app-summon-sidecar` and `app-secretless-broker` subcharts made in PR #346.

Resolves #[relevant GitHub issues, eg 76]

#### Change log

#### Test coverage

#### Documentation

#### Manual tests

### DIFF
--- a/bin/test-workflow/8_app_verify_authentication.sh
+++ b/bin/test-workflow/8_app_verify_authentication.sh
@@ -27,6 +27,7 @@ function finish {
     "INIT_WITH_HOST_OUTSIDE_APPS_PORT_FORWARD_PID"
     "SECRETLESS_PORT_FORWARD_PID"
     "SECRETS_PROVIDER_STANDALONE_PID"
+    "SECRETS_PROVIDER_INIT_PORT_FORWARD_PID"
   )
 
   # Upon error, dump some kubernetes resources and Conjur authentication policy
@@ -83,6 +84,7 @@ if [[ "$PLATFORM" == "openshift" ]]; then
   init_pod_with_host_outside_apps=$(get_pod_name test-app-with-host-outside-apps-branch-summon-init)
   secretless_pod=$(get_pod_name test-app-secretless)
   secrets_provider_standalone_pod=$(get_pod_name test-app-secrets-provider-standalone)
+  secrets_provider_init_pod=$(get_pod_name test-app-secrets-provider-init)
 
   # Routes are defined, but we need to do port-mapping to access them
   oc port-forward "$sidecar_pod" 8081:8080 > /dev/null 2>&1 &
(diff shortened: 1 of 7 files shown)


bin/test-workflow/8_app_verify_authentication.sh
```diff
@@ -104,6 +104,7 @@ else
   init_url_with_host_outside_apps="test-app-with-host-outside-apps-branch-summon-init.$TEST_APP_NAMESPACE_NAME.svc.cluster.local:8080"
   sidecar_url="test-app-summon-sidecar.$TEST_APP_NAMESPACE_NAME.svc.cluster.local:8080"
   secretless_url="test-app-secretless-broker.$TEST_APP_NAMESPACE_NAME.svc.cluster.local:8080"
+  secrets_provider_init_url="test-app-secrets-provider-init.$TEST_APP_NAMESPACE_NAME.svc.cluster.local:8080"
```
diverdane, line 107: We'll need a `secrets_provider_init_url` definition for `"$PLATFORM" == openshift" case above (Line 99).
We shouldn't need the `init_url` (Line 97) since that's not in our test matrix, and it's probably going to get deprecated real soon. So feel free to use port `8082` for `secrets_provider_init_url`.

diverdane (commented):
LGTM, except for missing port def.

bin/test-workflow/8_app_verify_authentication.sh
john-odonnell: Sounds good.
Does this deprecation also apply to `init_url_with_host_outside_apps`?

diverdane: Interesting question. I thought about bringing this up earlier, but I think this merits a discussion on whether this can be deleted or not. My thinking is that we should consider deleting this `init_url_with_host_outside_apps` test case. This test case is a bit of a corner case that's more about testing functionality of Conjur and Conjur security policy. All of the other application test cases use an `apps` policy group... and they require that the application Pod matches a specified Namespace, Deployment, ServiceAccount, and authn container name. For the `init_url_with_host_outside_apps`, the authn-k8s container uses a host ID that isn't part of the `apps` group, and it isn't as particular about checking Deployment and ServiceAccount.

For the other `secrets_provider_init_url` application, I looked back at the discussion for using `authn-k8s` as an init container. Apparently the consensus is not to deprecate, but to advise against using authn-k8s as an in it container  in production (i.e. advise using authn-k8s as a sidecar instead in production).

john-odonnell: Per our discussion - authn-k8s init container is not deprecated, but advised against in production - `init` URLs are maintained.

diverdane (approved):
Thanks for getting this workflow in!

diverdane (approved):

diverdane:
Thanks for adding the template fix!

PR #381, opened 2021-08-17
Add script for generating and diffing raw manifests

### LINKED ISSUE #370: Add GitHub Action that checks if raw manifests need to be generated from Helm charts
#350 Added a GHA which generated raw manifests from Helm charts with `helm template` and automatically committed the manifests when the charts changed. This change had to be reverted because it made use of the default `GITHUB_TOKEN` which had permissions to create commits in dev branches but did not have permissions to run subsequent GHA workflows. This meant that the other Helm test and code style enforcement Actions in the repository would not run afterwards. See [this section in the GHA docs](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#triggering-new-workflows-using-a-personal-access-token) for more details.

The new strategy is to add a GHA which generates temporary raw manifests and diffs them with the manifests which have already been committed. If there are changes then CI log will contain instructions for the developer to manually generate and commit the new manifests locally.

imheresamir:
### What does this PR do?
- Add GHA which generates temporary raw manifests and diffs them with the manifests which have already been committed. If there are changes then CI log will contain instructions for the developer to manually generate and commit the new manifests locally.

Resolves #370

#### Change log

#### Test coverage

#### Documentation

#### Manual tests


bin/helm-generate-raw-manifests-in-docker
```diff
@@ -0,0 +1,66 @@
...
+
+echo -e "\nDiffing generated/conjur-config-namespace-prep.yaml..."
+diff -cs helm/conjur-config-namespace-prep/generated/conjur-config-namespace-prep.yaml \
+            helm/conjur-config-namespace-prep/generated/conjur-config-namespace-prep.new.yaml
+
+if [[ "$diff_only" = false ]];
+then
+    echo -e "\nOverwriting old manifests..."
+
+    rm helm/conjur-config-cluster-prep/generated/conjur-config-cluster-prep.yaml
```
diverdane, line 59: I think this `rm` is not needed, the `mv` should be sufficient?

bin/helm-generate-raw-manifests-in-docker
```diff
@@ -0,0 +1,66 @@
...
+done
+
+echo "Building utility image with helm and yq..."
+docker build -f Dockerfile.helm-unit-test -t conjur-k8s-helm-generate-manifests:dev .
+
+echo "Generating raw k8s manifests..."
+docker run --name conjur-k8s-helm-generate-manifests --entrypoint "/conjur-authn-k8s-client/bin/helm-generate-raw-manifests" \
+            conjur-k8s-helm-generate-manifests:dev
+
+set +eo pipefail
```
diverdane, line 31: Do we need/want to ignore errors for the `docker cp ...` command?

If we really do want to ignore errors for this command, it might be easier to skip the `-o pipefail` setting at the top of the file (i.e. use `set -e`) since I don't think there is any piping used in the script. Then, for the `docker cp ...` commands here, you'd be able to use `<docker-cp-command> || true`.

bin/helm-generate-raw-manifests-in-docker
```diff
@@ -0,0 +1,66 @@
...
+diff -cs helm/conjur-config-cluster-prep/generated/conjur-config-cluster-prep.yaml \
+            helm/conjur-config-cluster-prep/generated/conjur-config-cluster-prep.new.yaml
+
+echo -e "\nDiffing generated/conjur-config-namespace-prep.yaml..."
+diff -cs helm/conjur-config-namespace-prep/generated/conjur-config-namespace-prep.yaml \
+            helm/conjur-config-namespace-prep/generated/conjur-config-namespace-prep.new.yaml
+
+if [[ "$diff_only" = false ]];
+then
+    echo -e "\nOverwriting old manifests..."
```
diverdane, line 57: If `$diff_only` is set to false, I think this will always echo the "Overwriting old manifests", even if they haven't changed. It might be nicer to check the results of the `diff -cs ...` commands above, and then do the overwrite only if the files have actually changed.

Another thought... If you do add checks for the results of the `diff -cs ...`, you should be able to remove the `If generated files have changed...` display above, and instead echo something more definitive here when the files have actually changed, e.g. `Generated files have changed. Please run ....`.

helm/conjur-config-cluster-prep/generated/conjur-config-cluster-prep.yaml
```diff
@@ -36,198 +36,12 @@ data:
...
   authnK8sAuthenticatorID: <Insert-Authenticator-ID-Here>
   authnK8sClusterRole: conjur-clusterrole
-  authnK8sNamespace: default
   authnK8sServiceAccount: conjur-serviceaccount
   # Conjur Configuration 
   conjurAccount: <Insert-Conjur-Account-Here>
   conjurApplianceUrl: https://insert.conjur.appliance.url.here
   conjurSslCertificate: "<Insert-Conjur-SSL-Certificate-Here>"
 ---
-# Source: conjur-config-cluster-prep/templates/tests/test-cluster-prep-configmap.yaml
```
diverdane, line 46: NICE cleanup!

diverdane (commented):
Looks pretty good!!!
A few questions/comments/suggestions.

bin/helm-generate-raw-manifests-in-docker
imheresamir: Done

bin/helm-generate-raw-manifests-in-docker
imheresamir: Refactored based on discussion

bin/helm-generate-raw-manifests-in-docker
imheresamir: Changed

bin/helm-generate-raw-manifests-in-docker
```diff
@@ -0,0 +1,75 @@
...
+  docker run --rm \
+    --name conjur-k8s-helm-generate-manifests \
+    -v "$REPO_ROOT:/conjur-authn-k8s-client" \
+    --entrypoint "/conjur-authn-k8s-client/bin/helm-generate-raw-manifests" \
+    conjur-k8s-helm-generate-manifests:dev
+
+  manifests_changed=false
+  for chart in "${HELM_CHARTS[@]}"; do
+    echo -e "\nDiffing generated/$chart.yaml..."
+    set +e
```
diverdane, line 45: It's possible to do this without temporarily setting `set +e`:
```
      ret_val=0
      diff -cs helm/$chart/generated/"$chart.old.yaml" helm/$chart/generated/"$chart.yaml" || ret_val=$?
```

diverdane (commented):
Looks even better!!!

bin/helm-generate-raw-manifests-in-docker
imheresamir: Good to know, thanks for the tip! Changed

diverdane (approved):
FANTASTIC!

PR #385, opened 2021-08-18
Implement E2E workflow for authenticator sidecars in OpenShift and Conjur Enterprise and follower in Jenkins

### LINKED ISSUE #245: There are end-to-end tests for Kubernetes sidecars with Conjur Enterprise and follower outside OpenShift
# Overview
CI automation is required to perform E2E testing of the deployment of authenticator sidecars
using Conjur Enterprise with followers outside of the OpenShift cluster in which the authenticator sidecars are deployed.
 
This task will use the E2E scripting that is created in Issue #239, and expanded to support deployment on OpenShift clusters with Issue #248, to deploy the authenticators and a sample sidecar application.
The scripts will assume that there is an OpenShift 4.6 cluster available to deploy authenticators and a sample application.

A "front-end" script will need to be included to spin up a Conjur Enterprise cluster with followers outside of the OpenShift cluster. This scripting will be the same as what is required for Issue #244.

# TBD

- __Which sidecar flows will run?__
~ Since E2E flows for all of the authenticator sidecar/init container types is already covered in
   Issue #242 (E2E using Conjur OSS), it may be sufficient to test one type of authenticator
   sidecar/init container here. (The assumption here is that if one authenticator sidecar/init
   container type can authenticate with a given configuration, the others "should" work in
   that they should be fairly agnostic of the endpoints they're using for authentication???) ~

Run all configurations for now to get things tested in OpenShift quickly. We will cut down on the test matrix as a separate effort.

# Dependencies
- Issue #228 (cluster prep Helm chart)
- Issue #237 (Namespace prep Helm chart)
- Issue #238 (sample application deployment Helm chart)
- Issue #239 (Reusable script for deploying applications and authenticators)
- Issue #247 (Add Secrets Provider support to sample application deployment Helm chart)
- Issue #248 (Add OpenShift support to reusable script for deploying applications and authenticators)

# DoD
- [ ] CI is able to spin up a Conjur cluster with follower(s) outside of the OpenShift cluster
- [ ] CI is able to successfully run the E2E authenticator / application test script

imheresamir:
### What does this PR do?
- Minor changes to get a new E2E flow working (authenticator sidecars in OpenShift and Conjur Enterprise and follower in Jenkins)
- Minor cleanups and deflaking

Resolves #245

#### Change log

#### Test coverage

#### Documentation

#### Manual tests

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -95,24 +95,37 @@ pipeline {
             sh './bin/helm-dependency-update-in-docker'
           }
         }
-        stage('Openshift E2E Workflow Tests') {
-          steps {
-            sh 'cd bin/test-workflow && summon --environment openshift -D ENV=ci -D VER=current ./start --platform oc'
-          }
-        }
-        stage('Run E2E Tests') {
+        stage('Test app with') {
           parallel {
-            stage('Enterprise and test app deployed to GKE') {
+            stage('Enterprise in GKE') {
               steps {
                 sh 'cd bin/test-workflow && summon --environment gke ./start --enterprise --platform gke'
               }
             }
-            stage('Enterprise deployed locally, test app deployed to GKE') {
+            stage('OSS in OpenShift') {
+              steps {
+                sh 'cd bin/test-workflow && summon --environment openshift -D ENV=ci -D VER=current ./start --platform openshift'
+              }
+            }
+          }
+        }
+        stage('Enterprise in Jenkins') {
+          stages {
+            stage('Test app in GKE') {
               steps {
                 sh '''
                   HOST_IP="$(curl http://169.254.169.254/latest/meta-data/public-ipv4)";
                   echo "HOST_IP=${HOST_IP}"
                   cd bin/test-workflow && summon --environment gke ./start --enterprise --platform jenkins
+                '''
+              }
+            }
+            stage('Test app in OpenShift') {
+              steps {
+                sh '''
+                  HOST_IP="$(curl http://169.254.169.254/latest/meta-data/public-ipv4)";
+                  echo "HOST_IP=${HOST_IP}"
+                  cd bin/test-workflow && summon --environment openshift -D ENV=ci -D VER=current ./start --enterprise --platform jenkins
                 '''
               }
             }
@@ -125,9 +138,10 @@ pipeline {
       parallel {
         stage('On a master build') {
           when { branch 'master' }
-            steps {
-              sh 'summon ./bin/publish --edge'
-            }
+
+          steps {
+            sh 'summon ./bin/publish --edge'
+          }
         }
         stage('On a new tag') {
           when { tag "v*" }
--- a/bin/test-workflow/0_prep_env.sh
+++ b/bin/test-workflow/0_prep_env.sh
@@ -6,9 +6,9 @@ set -o pipefail
 export CONJUR_OSS_HELM_INSTALLED="${CONJUR_OSS_HELM_INSTALLED:-true}"
 export UNIQUE_TEST_ID="$(uuidgen | tr "[:upper:]" "[:lower:]" | head -c 10)"
 
-# PLATFORM is used to differentiate between general Kubernetes platforms (K8s vs. oc), while
-# CONJUR_PLATFORM is used to differentiate between sub-platforms (for vanilla K8s, KinD vs. GKE) for the Conjur deployment
-# APP_PLATFORM serves the same purpose as CONJUR_PLATFORM, but for the test app deployment
+# PLATFORM is used to differentiate between general Kubernetes platforms (kubernetes, openshift), while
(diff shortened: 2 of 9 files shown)


bin/test-workflow/stop
```diff
@@ -8,31 +8,49 @@ check_env_var RUN_CLIENT_CONTAINER
...
   "
-  rm -rf temp/kubernetes-conjur-deploy-"$UNIQUE_TEST_ID"
+
+  rm -rf "temp/kubernetes-conjur-deploy-$UNIQUE_TEST_ID"
 
 elif [[ "$CONJUR_PLATFORM" == "jenkins" ]]; then
-  pushd temp/conjur-intro-"$UNIQUE_TEST_ID" > /dev/null
+  pushd "temp/conjur-intro-$UNIQUE_TEST_ID" > /dev/null
     ./bin/dap --stop
+    docker run --rm -w /dap-intro -v "$(pwd):/dap-intro" alpine ash -c "rm -rf system"
```
john-odonnell, line 42: Just a question here: what is this command removing? What is is necessary?

bin/test-workflow/0_prep_env.sh
```diff
@@ -6,30 +6,37 @@ set -o pipefail
...
 export CONJUR_PLATFORM
 
-if [[ "$CONJUR_PLATFORM" == "oc" ]]; then
-  PLATFORM="openshift"
+if [[ "$CONJUR_PLATFORM" == "openshift" ]]; then
+  PLATFORM="${PLATFORM:-openshift}"
 else
-  PLATFORM="kubernetes"
+  PLATFORM="${PLATFORM:-kubernetes}"
 fi
```
john-odonnell, line 23: In these cases, do we want to defer to the existing `PLATFORM` variable? Might allow for `CONJUR_PLATFORM` and `PLATFORM` to be out of phase, instead of aligning them.

Jenkinsfile
```diff
@@ -95,19 +95,23 @@ pipeline {
...
+            stage('OSS in OpenShift') {
+              steps {
+                sh 'cd bin/test-workflow && summon --environment openshift -D ENV=ci -D VER=current ./start --platform openshift'
+              }
+            }
+          }
+        }
+        stage('Enterprise in Jenkins') {
+          stages {
+            stage('Test app in GKE') {
```
john-odonnell, line 114: Could we run these stages in parallel?
If not, I think they probably compete to deploy Enterprise to the same port. Wondering if we could deploy one Conjur instance, and have parallel stages authenticate with the same Conjur instance. Not sure if that would even work, just a thought, and definitely out of scope.

Jenkinsfile
```diff
@@ -95,19 +95,23 @@ pipeline {
             sh './bin/helm-dependency-update-in-docker'
           }
         }
-        stage('Openshift E2E Workflow Tests') {
-          steps {
-            sh 'cd bin/test-workflow && summon --environment openshift -D ENV=ci -D VER=current ./start --platform oc'
-          }
-        }
-        stage('Run E2E Tests') {
+        stage('Test app with') {
```
john-odonnell, line 98: nit: I think this stage should have a title similar to `Enterprise in Jenkins`, maybe something like `Conjur in K8s Cloud`? `Test app with` feels too nondescript.

john-odonnell (changes requested):
Looks good! Two suggestions, two questions:

Jenkinsfile
john-odonnell: "These stages" referring to `Test app in GKE` and `Test app in Openshift`

bin/test-workflow/0_prep_env.sh
imheresamir: That's exactly why the change was made - for the flow being added `PLATFORM=openshift` and `CONJUR_PLATFORM=jenkins`

Jenkinsfile
imheresamir: Two Enterprise installs in the same Jenkins agent clobber each other when run in parallel. A way around this would be to do the Enterprise installs in containers with container listening ports being forwarded to different Jenkins agent listening ports - and then configuring the authenticators to talk to Enterprise on these ports - but I think its more trouble than its worth.

imheresamir: A single Enterprise instance with the policy loading involved to support multiple authenticators and test apps sounds pretty difficult to troubleshoot and not worth it for what we're trying to achieve here IMO - I think we want to keep these E2E flows simple and easy to debug even at the expense of some parallelism

john-odonnell: Good point, and you're right - more trouble than it's worth.

Jenkinsfile
john-odonnell: After looking at how the pipelines displayed graphically - I think this change works. I was evaluating the Jenkinsfile naming without considering it's output.

john-odonnell (approved):
Answered all my comments and questions! LGTM 😃

PR #387, opened 2021-08-23
Runs test apps in E2E workflow as non-root user

diverdane:
### What does this PR do?

This change modifies the deployment manifests for applications in the E2E workflow so that the test application runs as a non-root user. Running containers using non-root user/group is considered best practice for Kubernetes workloads.

This is accomplished by setting the associated 'PodSecurityContext' for each deployment so that every container (i.e. the application container as well as the authentication container, if present) are run using:

- The 'nobody' non-root user (UserID 66634)
- The 'nobody' non-root group (GroupID 65534)
- fsGroup of 65534 (i.e. directories and files for all shared volumes
                    are created using a Group ID of 65534).

In some cases, this also requires creating `volumeMounts` for containers and `emptyDir` Volumes for the Pods for directories to which the authenticator client containers write, or from which application containers will read. This is required so that these directories and their contents will be created with the `65534` GroupID, as is needed since the containers will now be running as User:group of `65534:65534`. Without these `emptyDir` additions, the authenticator containers will lack permission to write files to these directories, and the application container will lack permission to read from these directories.

#### Change log

#### Test coverage

#### Documentation

#### Manual tests

### DIFF
--- a/helm/conjur-app-deploy/charts/app-secretless-broker/templates/test_app_secretless_broker.yaml
+++ b/helm/conjur-app-deploy/charts/app-secretless-broker/templates/test_app_secretless_broker.yaml
@@ -80,11 +80,25 @@ spec:
           mountPath: /etc/secretless
           name: config
           readOnly: true
+        - name: conjur-access-token
+          mountPath: /run/conjur
+        - name: conjur-certs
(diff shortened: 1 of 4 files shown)


imheresamir (approved):
LGTM

PR #397, opened 2021-09-02
Uses just-built test client image for E2E CI

diverdane:
### What does this PR do?

This fixes the E2E Jenkins CI tests so that they use a just-built test client image (which is used for kubectl, oc, helm, openssl, etc. clients) when running E2E test scripts, rather than using the `edge` version of this container that had been published on the PREVIOUS merge to the master branch.

Without this fix, there could potentially be flaws in the current, just-built client container image, but the CI test run will not catch the issue. If this occurs e.g. on a merge to the master branch, then the flawed, just-built client container image will be published as the new `edge` version, and all subsequent test runs will be broken.

For anyone using the E2E test scripts locally/manually, the version tag for the test client image defaults to `edge`, so that our E2E scripts are not dependent upon having to do a build of container images. This is convenient for those who may want to use the E2E scripts to explore Conjur authentication.

#### Change log

#### Test coverage

#### Documentation

#### Manual tests

### DIFF
--- a/bin/test-workflow/4_admin_cluster_prep.sh
+++ b/bin/test-workflow/4_admin_cluster_prep.sh
@@ -4,8 +4,10 @@ set -euo pipefail
 cd "$(dirname "$0")" || ( echo "cannot cd into dir" && exit 1 )
 
 TIMEOUT="${TIMEOUT:-5m0s}"
+TEST_CLIENT_IMAGE_TAG="${TEST_CLIENT_IMAGE_TAG:-edge}"
 
 source utils.sh
+source ../build_utils
 
 check_env_var CONJUR_APPLIANCE_URL
 check_env_var CONJUR_NAMESPACE_NAME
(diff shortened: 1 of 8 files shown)


bin/test-workflow/4_admin_cluster_prep.sh
```diff
@@ -4,6 +4,8 @@ set -euo pipefail
 cd "$(dirname "$0")" || ( echo "cannot cd into dir" && exit 1 )
 
 TIMEOUT="${TIMEOUT:-5m0s}"
+TEST_CLIENT_IMAGE_TAG="${TEST_CLIENT_IMAGE_TAG:-edge}"
+echo "***TEMP*** env: TEST_CLIENT_IMAGE_TAG: $TEST_CLIENT_IMAGE_TAG"
```
rpothier, line 8: Was this intended to stay in?

rpothier (approved):
LGTM! Just one minor comment.

bin/test-workflow/4_admin_cluster_prep.sh
diverdane: No, this is some temporary debug code. I think I figured out what was missing, so I'll remove this now.

PR #398, opened 2021-09-09
Bump version to 0.21.1

jtuttle:


CHANGELOG.md
```diff
@@ -166,11 +168,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Added
 - Added a RedHat-certified version of the image.
 
-## [0.10.1] - 0000-00-00
+## 0.10.1 - 0000-00-00
```
orenbm, line 171: why were the brackets removed?

jtuttle: The brackets are only necessary if there's a link specified somewhere in the file. Most of our projects don't include a link for the first version.

jtuttle: Ah I see you added this rule @orenbm : https://github.com/cyberark/parse-a-changelog/blob/main/lib/grammar.tt#L83

Wouldn't it make more sense to link to a list of commits? For example, something like: https://github.com/cyberark/conjur-authn-k8s-client/commits/v0.10.1

This is much more similar to what the other links in the file do, i.e. show a list of commits that's included in the released tag.

orenbm: i may have not been aware to the fact that we don't include a link to the first version. feel free to remove the constrain i added and go in another path.

Thanks!

diverdane (approved):
LGTM!!!

PR #425, opened 2021-12-15
Refactor authn-k8s-client to be authentication flow generic

tzheleznyak:
### Desired Outcome

As part of the authn-jwt side car in K8S this is first stage where we do refactor, that will allow adding more authentication methods.

Defining interface for Config And Authenticator and factories creating instances of them.

### Connected Issue/Story

ONYX-14719

CyberArk internal issue link: [insert issue ID]()

### Definition of Done
*At least 1 todo must be completed in the sections below for the PR to be
merged.*

#### Changelog

  CHANGELOG update

#### Test coverage

  changes, or

#### Documentation

#### Behavior

- [x] This PR changes product behavior and has been reviewed by a PO, or

#### Security


pkg/log/log_messages.go
```diff
@@ -75,3 +75,5 @@ const CAKC059 string = "CAKC059 Path exists but does not contain regular file: %
 const CAKC060 string = "CAKC060 Setting %s given invalid value %s"
 const CAKC061 string = "CAKC061 Failed to validate setting for Authenticator configuration"
 const CAKC062 string = "CAKC062 Required Authenticator setting %s not provided"
+const CAKC063 string = "CAKC63 Can't find configuration for this authn URL"
+const CAKC064 string = "CAKC64 Can't find Authenticator for this configuration"
```
sashaCher, line 79:
```suggestion
const CAKC063 string = "CAKC063 Can't find configuration for this authn URL"
const CAKC064 string = "CAKC064 Can't find Authenticator for this configuration"
```

pkg/log/log_messages.go
```diff
@@ -75,3 +75,5 @@ const CAKC059 string = "CAKC059 Path exists but does not contain regular file: %
 const CAKC060 string = "CAKC060 Setting %s given invalid value %s"
 const CAKC061 string = "CAKC061 Failed to validate setting for Authenticator configuration"
 const CAKC062 string = "CAKC062 Required Authenticator setting %s not provided"
+const CAKC063 string = "CAKC063 Can't find configuration for this authn URL"
```
nessiLahav, line 78: Add TW here

nessiLahav (approved):
LGTM

pkg/log/log_messages.go
```diff
@@ -75,3 +75,5 @@ const CAKC059 string = "CAKC059 Path exists but does not contain regular file: %
 const CAKC060 string = "CAKC060 Setting %s given invalid value %s"
 const CAKC061 string = "CAKC061 Failed to validate setting for Authenticator configuration"
 const CAKC062 string = "CAKC062 Required Authenticator setting %s not provided"
+const CAKC063 string = "CAKC063 Can't find configuration for URL : %s"
```
tzheleznyak, line 78: @shulifink

tzheleznyak: May you review this two log messages please?

shulifink, line 78:
```suggestion
const CAKC063 string = "CAKC063 Can't find configuration for URL: %s"
```

pkg/log/log_messages.go
```diff
@@ -75,3 +75,5 @@ const CAKC059 string = "CAKC059 Path exists but does not contain regular file: %
...
+const CAKC064 string = "CAKC064 Can't find Authenticator for this configuration"
```
shulifink, line 79: @tzheleznyak  - What is the use-case for this?

shulifink (changes requested):

diverdane:
@tzheleznyak,

Thanks for taking on this challenge! I think the interfaces that you've defined are a good start at making this client more pluggable. I'd like to make some suggestions that might simplify things a little bit. This will take a bit of rearranging, so we can chat about this tomorrow, and maybe pull some other folks in to get some consensus (Kumbi and Johah might be good to include).

First, some high-level comments:

1. With the proposed directory structure, there seems to be a "mix of concerns" in the `common`
   and `creators` directory. For example, each of these directories contains code that does
   configuration work as well as code that does authentication work. Another way of looking at it,
   the configuration code is split out between the `common` and `creators` directories. (I'm
   thinking that this mix of concerns might have caused the cyclic dependency references???)
   I think it might be cleaner to organize things similar to the original directory structure:
   ```
   -pkg/
      - authenticator/     # Common authentication code
         - config/               # Common config code
         - k8s/                   # "Concrete" instantiations of config/authen interfaces for K8s
         - jwt/                     # "Concrete" instantiations of config/authen interfaces for JWT
   ```

2. I see that for each authenticator type (authn-k8s or authn-jwt), you're defining a config structure and an
   authenticator structure, with each structure satisfying the `ConfigurationInterface` and the
   `AuthenticatorInterface` interfaces respectively. I'm thinking that it might simplify things to have one combined
   "Authenticator" structure for each authenticator type. That structure can satisfy both
   `ConfigurationInterface` and `AuthenticatorInterface` interfaces. Going even further, we could
   potentially combine both interfaces into a single "Authenticator" interface, where each concrete
   realization of that interface knows e.g. what config variables are required, what defaults to use for
   config variables, and how to initialize and authenticate with Conjur. This would essentially combine
   everything we need to know about an authenticator. Once we create a structure that satisfies
   the single, "Authenticator" interface, then it's pretty easy and clean to pass it around to do
   configuration and authentication actions. We wouldn't have to do type switches/assertions like the following:
   ```
   var cfg = (*config).(*Config)
   ```
   I think there are some methods that can be dropped from the interface definitions, so that a
   combined interface might look like this:
   ```
   type Authenticator interface {                                         
     LoadConfig(settings map[string]string)                                                                                 
     GetEnvVariables() []string                                                  
     GetRequiredVariables() []string                                             
     GetDefaultValues() map[string]string                                        
     Init(config *ConfigurationInterface) (AuthenticatorInterface, error)                                                                                                                    
     Authenticate() error                                                                                          
 }

pkg/authenticator/common/configuration_interface.go
```diff
@@ -0,0 +1,16 @@
...
+	"time"
+)
+
+type ConfigurationInterface interface {
+	LoadConfig(settings map[string]string)
+	GetAuthenticationType() string
+	GetEnvVariables() []string
+	GetRequiredVariables() []string
+	GetDefaultValues() map[string]string
+	GetContainerMode() string
```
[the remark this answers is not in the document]
tzheleznyak: These are all configuration details so they are related to configuration in my opinion

tzheleznyak:
@diverdane 
About 1:

During development i came across dependency loop when i tried putting interfaces and factories in the same package. It what made me split factories to one separate and the interfaces to the common package. 

The common package is all code relevant to the main authenticator package that don't uses other packages inside authenticator and used by multiple packages inside authenticator.

The name creators can be discussed and i can remove this directory and move the factories one step back.


About 2:
I don't think we need to combine to one entity the ConfigInterface and AuthenticatorInterface. They have separate resnponsibilits . Config responsible for:
* Collecting Config Data - Currently from env but it future may be from future sources like DB/network and authenticator shouldn't care where it comes from
* Validating the config data
* Storing the config data

These are not responsibilits of the authenticator that should only support two things:
* Authenticate
* Providing access token

pkg/log/log_messages.go
tzheleznyak: We are going to add support to authn-jwt to authn-k8s-client.
And this is the error message when the authn-client can't decide which authentication flow should run (authn-k8s or authn-jwt)

pkg/authenticator/creators/authenticator_factory.go
```diff
@@ -0,0 +1,36 @@
...
+	var error error
+	var authn common.AuthenticatorInterface
+	authn, error = GetAuthenticator(conf)
+	if error != nil {
+		return nil, error
+	}
+	return authn.InitWithAccessToken(&conf, token)
+}
+
+func GetAuthenticator(conf common.ConfigurationInterface) (common.AuthenticatorInterface, error) {
```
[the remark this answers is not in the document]
tzheleznyak: Good catch

pkg/log/log_messages.go
shulifink: How about "No authenticator configured. "

shulifink: or even better: "No authentication configured. "

pkg/authenticator/common/configuration_interface.go
```diff
@@ -0,0 +1,16 @@
+package common
+
+import (
+	"time"
+)
+
+type ConfigurationInterface interface {
+	LoadConfig(settings map[string]string)
+	GetAuthenticationType() string
```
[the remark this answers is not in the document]
tzheleznyak: Talked about not merging authenticator and config

pkg/log/log_messages.go
```diff
@@ -75,3 +75,5 @@ const CAKC059 string = "CAKC059 Path exists but does not contain regular file: %
 const CAKC060 string = "CAKC060 Setting %s given invalid value %s"
 const CAKC061 string = "CAKC061 Failed to validate setting for Authenticator configuration"
 const CAKC062 string = "CAKC062 Required Authenticator setting %s not provided"
+const CAKC063 string = "CAKC063 Can't find configuration for URL: %s"
```
doodlesbykumbi, line 78: In other repos we have been standardising on `Unable`. So these would all become `Unable to ...`

pkg/authenticator/k8s/config.go
```diff
@@ -0,0 +1,133 @@
...
+	return envVariables
+}
+
+func (config *Config) GetRequiredVariables() []string {
+	return requiredEnvVariables
+}
+
+func (config *Config) GetDefaultValues() map[string]string {
+	return defaultValues
+}
```
doodlesbykumbi, line 121: These methods are a bit strange because they don't make use of the struct at all, and so could be static.

pkg/authenticator/k8s/authenticator.go
```diff
@@ -48,9 +48,27 @@ const (
...
+	log.Debug(log.CAKC058)
+	var cfg = (*config).(*Config)
+
+	return NewWithAccessToken(*cfg, token)
+}
+
 // New creates a new authenticator instance from a token file
-func New(config authnConfig.Config) (*Authenticator, error) {
-	accessToken, err := file.NewAccessToken(config.TokenFilePath)
+func New(config Config) (*Authenticator, error) {
```
doodlesbykumbi, line 70: I think it would be good for us to update this to `NewAuthenticator` to make it explicit what this method is creating.

pkg/authenticator/k8s/authenticator.go
```diff
@@ -48,9 +48,27 @@ const (
...
+	var cfg = (*config).(*Config)
+	authn, err := New(*cfg)
+	if err != nil {
+		return nil, fmt.Errorf(log.CAKC019)
+	}
+
+	return authn, nil
+}
+
+func (auth *Authenticator) InitWithAccessToken(config *common.ConfigurationInterface, token access_token.AccessToken) (common.AuthenticatorInterface, error) {
```
doodlesbykumbi, line 62: I'm not sure about the name of the method. I think we need something other than `Init` that better captures what this method is doing.

This should also probably be a free standing function (not a method) since it doesn't use the auth struct at all.

doodlesbykumbi (commented):
Some thoughts

pkg/authenticator/k8s/config.go
tzheleznyak: But the usage inside Validate() and GatherSettings() 
Is config.GetDefaultValues() so it will go to the config of the correct flow

pkg/authenticator/k8s/authenticator.go
tzheleznyak: Done

pkg/authenticator/k8s/authenticator.go
tzheleznyak: Lets discuss in the meeting

pkg/authenticator/common/configuration_interface.go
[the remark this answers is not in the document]
diverdane: Yes, this makes sense to leave separate, thanks!

pkg/authenticator/common/configuration_interface.go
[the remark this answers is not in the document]
diverdane: Right. My original (not completely thought-through) idea was whether we could remove them from the interface and add them as common functions in the config (a.k.a. common) directory. But I see now why these accessor functions are needed in this interface.

pkg/authenticator/authenticator_factory.go
```diff
@@ -0,0 +1,30 @@
...
+import (
+	"fmt"
+	"github.com/cyberark/conjur-authn-k8s-client/pkg/access_token"
+	"github.com/cyberark/conjur-authn-k8s-client/pkg/access_token/file"
+	"github.com/cyberark/conjur-authn-k8s-client/pkg/authenticator/config"
+	k8sAuthenitcator "github.com/cyberark/conjur-authn-k8s-client/pkg/authenticator/k8s"
+	"github.com/cyberark/conjur-authn-k8s-client/pkg/log"
+)
+
+func NewAuthenticator(conf config.Configuration) (Authenticator, error) {
```
diverdane, line 12: I don't want to hold up the merge on this, but Go style guides (e.g. https://go.dev/doc/effective_go#commentary) recommend that any exported names should have a preceding doc comment that begins with that name, e.g. 
```
// NewAuthenticator creates an instance of the Authenticator interface based on configured authenticator type.
```
(My IDE is running `golint` automatically, so when I write a file I see warnings pop up. BTW `golint` is deprecated, but there are replacements, e.g. `golangci-lint`.)

diverdane (commented):
This looks FANTASTIC!!! Thanks again for your patience on this! I really like how you broke out things into `config` and `common`.
I just have a few very minor editorial nits, otherwise this looks good-to-go.

diverdane (approved):
LGTM!!!

nessiLahav (approved):
LGTM

alexkalish:
Internal issue: [ONYX-14722](https://ca-il-jira.il.cyber-ark.com:8443/browse/ONYX-14722).

PR #426, opened 2021-12-23
Add JWT flow

tzheleznyak:
### Desired Outcome

Added JWT flow to the authenticator

### Connected Issue/Story

ONYX-14722

#### Changelog

  CHANGELOG update

#### Test coverage

  changes, or

#### Documentation

#### Behavior

#### Security

- [ ] There are no security aspects to these changes


pkg/authenticator/jwt/authenticator.go
```diff
@@ -0,0 +1,130 @@
...
+)
+
+// Authenticator contains the configuration and client
+// for the authentication connection to Conjur
+type Authenticator struct {
+	client      *http.Client
+	privateKey  *rsa.PrivateKey
+	AccessToken access_token.AccessToken
+	Config      *Config
+	PublicCert  *x509.Certificate
```
doodlesbykumbi, line 27: Any reason for making these public ? I also realised that the Authenticator interface should expose `AccessToken` through some getter. I think we spoke about having Authenticate and GetAccessToken on the call.

doodlesbykumbi (commented):
Some comments

pkg/authenticator/config/validate_test.go
```diff
@@ -37,15 +37,49 @@ func TestValidate(t *testing.T) {
 			},
 			assert: assertEmptyErrorList(),
 		},
+		{
+			description: "happy path - jwt",
```
tzheleznyak, line 41: Added some UT for env variables validation @eladkug

pkg/authenticator/jwt/requests.go
```diff
@@ -0,0 +1,39 @@
...
+	formattedJwt := fmt.Sprintf("jwt=%s", jwtToken)
+	requestBody := strings.NewReader(formattedJwt)
+
+	if req, err = http.NewRequest("POST", authenticateURL, requestBody); err != nil {
+		return nil, log.RecordedError(log.CAKC023, err)
+	}
+
+	req.Header.Set("Content-Type", "text/plain")
+	req.Header.Set("Content-Length", string(len(formattedJwt)))
+	req.Header.Set("User-Agent", "k8s")
```
sashaCher, line 36: Do we want to add support for [accept-encoding](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Encoding) header?
Conjur is able to return access token already encoded base64 instead of plain json ([1](https://github.com/cyberark/conjur/blob/27f2107987f1fceb73d6ebbbc3fb37bcba6aec90/app/controllers/authenticate_controller.rb#L217), [2](https://github.com/cyberark/conjur/blob/27f2107987f1fceb73d6ebbbc3fb37bcba6aec90/app/controllers/authenticate_controller.rb#L336))

pkg/authenticator/config/validate_test.go
eladkug: why you removed filepath and update host?

pkg/authenticator/jwt/requests.go
tzheleznyak: Currently the components that use authn-k8s-client do the encode themselves . I think this is good idea but not in scope. I wanted to be aligned with the auhtn-k8s flow

pkg/authenticator/config/validate_test.go
tzheleznyak: Because until now there was only k8s authentication so we could put random string in UT like file-path


And now when we have a factory that check if string includes authn-k8s or authn-jwt to decide authentication flow

pkg/authenticator/jwt/authenticator.go
tzheleznyak: Changes to non public.

But don't see why we need getAccessToken didn't found usage of authn.AccessToken in secretless and in provider.
tell me if im wrong and i will add it back

CHANGELOG.md
```diff
@@ -8,10 +8,13 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Added
 - Add support for tracing with OpenTelemetry. This adds a new function to the authenticator, `AuthenticateWithContext`. The existing funtion, `Authenticate()` is deprecated and will be removed in a future upddate. [cyberark/conjur-authn-k8s-client#423](https://github.com/cyberark/conjur-authn-k8s-client/pull/423)
+- Authn JWT flow support [cyberark/conjur-authn-k8s-client#426](https://github.com/cyberark/conjur-authn-k8s-client/pull/426)
```
tzheleznyak, line 11: @szamir1 @shulifink 
May you review my code

pkg/authenticator/jwt/requests.go
```diff
@@ -0,0 +1,39 @@
...
+	"strings"
+
+	"github.com/cyberark/conjur-authn-k8s-client/pkg/log"
+)
+
+// AuthenticateRequest sends an authenticate request
+func AuthenticateRequest(authnURL string, account string, username string, jwtToken string) (*http.Request, error) {
+	var authenticateURL string
+
+	if len(username) > 0 {
```
nessiLahav, line 16: we can extract this logic to other function, and is will be more easy to UT it

pkg/authenticator/jwt/authenticator.go
```diff
@@ -0,0 +1,133 @@
...
+		return nil, err
+	}
+
+	if auth.Config.Common.Username != nil {
+		authenticatingIdentity = auth.Config.Common.Username.FullUsername
+	} else {
+		authenticatingIdentity = ""
+	}
+
+	req, err := AuthenticateRequest(
```
nessiLahav, line 105: i think we should align the context handling with this [code](https://github.com/cyberark/conjur-authn-k8s-client/blob/b5ff74862c3a8fcd09be6763472289e006bda5c2/pkg/authenticator/k8s/authenticator.go#L309)

nessiLahav (changes requested):
Small comments, great job

pkg/authenticator/jwt/authenticator.go
tzheleznyak: This what we talked about and i fixed

tzheleznyak: Oh i see i done it in all error handling in the function

pkg/authenticator/jwt/authenticator.go
```diff
@@ -0,0 +1,140 @@
...
+	authenticationResponse, err := auth.sendAuthenticationRequest(spanCtx, tr)
+	if err != nil {
+		span.RecordErrorAndSetStatus(err)
+		span.End()
+		return err
+	}
+
+	err = auth.accessToken.Write(authenticationResponse)
+	if err != nil {
+		return err
```
szh, line 77: This should have
```go
span.RecordErrorAndSetStatus(err)
```

pkg/authenticator/jwt/authenticator.go
```diff
@@ -0,0 +1,140 @@
...
+func (auth *Authenticator) sendAuthenticationRequest(ctx context.Context, tracer trace.Tracer) ([]byte, error) {
+	var authenticatingIdentity string
+
+	_, span := tracer.Start(ctx, "Send authentication request")
+	defer span.End()
+
+	jwtToken, err := loadJWTToken(auth.Config.JWTTokenFilePath)
+
+	if err != nil {
+		return nil, err
```
szh, line 96: This should have
```go
span.RecordErrorAndSetStatus(err)
```

pkg/authenticator/jwt/tests/authenticator_test.go
```diff
@@ -0,0 +1,121 @@
...
+			name:         "wrong url given",
+			jwtTokenPath: tmpJwtTokenPath,
+			assert: func(t *testing.T, authn *jwt.Authenticator, err error) {
+				assert.NotNil(t, err)
+				assert.True(t, strings.Contains(err.Error(), "Failed to send https authenticate request or receive response"))
+			},
+			wrongUrl: true,
+		},
+		{
+			name: "token not exists",
```
szh, line 54:
```suggestion
			name: "token doesn't exist",
```

pkg/authenticator/jwt/tests/authenticator_test_server.go
```diff
@@ -0,0 +1,48 @@
...
+
+import (
+	"crypto/tls"
+	"crypto/x509"
+	"net/http"
+	"net/http/httptest"
+	"strings"
+)
+
+type TestAuthServer struct {
```
szh, line 11: This file seems to be a slimmed down version of `pkg/authenticator/k8s/tests/authenticator_test_server.go`. Can we move that a location where it can be reused instead of duplicated?

szh (changes requested):
Looking good, left a few comments

pkg/authenticator/jwt/tests/authenticator_test_server.go
tzheleznyak: Good point thanks

szh (approved):
LGTM! 👍

pkg/authenticator/config/configuration_factory.go
```diff
@@ -44,6 +45,8 @@ func ConfigFromEnv(readFileFunc common.ReadFileFunc) (Configuration, error) {
 func getConfiguration(url string) (Configuration, error) {
 	if strings.Contains(url, k8sAuthenticator.AuthnType) {
 		return &k8sAuthenticator.Config{}, nil
+	} else if strings.Contains(url, jwtAuthenticator.AuthnType) {
```
diverdane, line 48: Instead of using an `if/else` block here, you could optionally use `switch` without a statement (i.e. evaluating expressions in the case statements). I think this makes it a little more obvious that there are 3 possible results for which we're looking. For example:
```
	switch {
	case strings.Contains(url, k8sAuthenticator.AuthnType):
		return &k8sAuthenticator.Config{}, nil
	case strings.Contains(url, jwtAuthenticator.AuthnType):
		return &jwtAuthenticator.Config{}, nil
	default:
		return nil, fmt.Errorf(log.CAKC063, url)
	}
```

pkg/authenticator/jwt/config.go
```diff
@@ -0,0 +1,110 @@
...
+	"CONJUR_CLIENT_CERT_PATH":              DefaultClientCertPath,
+	"CONJUR_AUTHN_TOKEN_FILE":              DefaultTokenFilePath,
+	"CONJUR_TOKEN_TIMEOUT":                 DefaultTokenRefreshTimeout,
+	"CONJUR_CLIENT_CERT_RETRY_COUNT_LIMIT": DefaultClientCertRetryCountLimit,
+	"JWT_TOKEN_PATH":                       DefaultJWTTokenPath,
+}
+
+func configureLogLevel(level string) {
+	validVal := "true"
+	if level == validVal {
```
diverdane, line 62: You could also use a switch statement here. I think this makes it a little more obvious that there are three possible outcomes (2 good, 1 bad):
```
	switch level
	case validVal:
		log.EnableDebugMode()
	case "":
		// Log level not configured
		break
	default:
		// Log level is configured but it's invalid
		log.Warn(log.CAKC034, level, validVal)
	}
```

diverdane (dismissed):
Looks good!!!
Some editorial comments and a possible missing `strconv.Itoa()`.

nessiLahav (dismissed):

pkg/authenticator/jwt/config.go
```diff
@@ -0,0 +1,110 @@
...
+	}
+}
+
+func (config *Config) LoadConfig(settings map[string]string) {
+	configureLogLevel(settings["DEBUG"])
+
+	config.Common = common.Config{}
+	config.Common.LoadConfig(settings)
+
+	for key, value := range settings {
```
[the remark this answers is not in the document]
tzheleznyak: Good point thanks

pkg/authenticator/jwt/config.go
tzheleznyak: Good point. Moved it to ConfigFactory to be in one place

nessiLahav (approved):
LGTM

PR #431, opened 2022-01-03
Adds file format annotation for Secrets Provider P2F helm chart

diverdane:
### Desired Outcome

When the E2E test scripts are run, including the deployment of an application that uses the Secrets Provider as an Init container in Push-to-File mode, and then that Deployment is modified to use a "freshly-built" Secrets Provider image (i.e. image built using most recent code from Secrets Provider branch), then the Pet Store application should successfully come up.

Currently, when a "freshly-built" Secrets Provider image is used, the Pet Store application fails because the secret file that is created is defaulting to YAML format (because the `secret-file-format.<secret-group>` annotation is not explicitly set to `template`). In this scenario, the secret file needs to be a Spring boot application format file, built using the custom template supplied via Annotation.

### Implemented Changes

Added an Annotation to explicitly set the secret file format to `template` for the corresponding Helm subchart.

### Definition of Done

- [x] When the `bin/test-workflow` scripts are run, and a freshly-built Secrets Provider image is used for the deployment that uses SP in Init container mode running in push-to-file mode, the application successfully comes up.

#### Changelog

- [x] This PR does not include user-facing changes and doesn't require a
  CHANGELOG update

#### Test coverage

  changes, or

#### Documentation

#### Behavior

- [x] No behavior was changed with this PR

#### Security

### DIFF
--- a/helm/conjur-app-deploy/charts/app-secrets-provider-p2f/templates/test_app_secrets_provider_p2f.yaml
+++ b/helm/conjur-app-deploy/charts/app-secrets-provider-p2f/templates/test_app_secrets_provider_p2f.yaml
@@ -43,6 +43,7 @@ spec:
           - test-secrets-provider-p2f-app-db/username
           - test-secrets-provider-p2f-app-db/password
         conjur.org/secret-file-path.p2f-app: "./application.yaml"
+        conjur.org/secret-file-format.p2f-app: template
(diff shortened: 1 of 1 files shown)


john-odonnell (approved):
LGTM!

szh (approved):

PR #434, opened 2022-01-04
Remove pattern check in schema validation for cert file path

rpothier:
The conjur-config-cluster-prep Helm schema tests check that the certificate file path
has `files` or `tests` in the path. The file path is not restricted
to those directories and it should allow other paths.

### Desired Outcome

The user should be able to specify an alternative path for the certificate file.
The files directory is only the default for the `get-conjur-cert.sh`
script but it is not required to be in this directory and the script is not required to be used,
this is only a helper script, the user can use an alternate method to get the cert.
Note that the certificate must be located inside the chart as Helm does not have access to
files outside the chart.

### Implemented Changes

Removed the check for `files` in the schema verification.

Also added a change to bin/test-workflow/0_prep_env.sh
to use an existing UNIQUE_TEST_ID if it exists. This is only for debugging and allows
the user to set this variable and then run ./start -n and ./stop independently.
Jenkins will not have this set and should create a new unique id each time. 

### Connected Issue/Story

CyberArk internal issue link: [ONYX-15540](https://ca-il-jira.il.cyber-ark.com:8443/browse/ONYX-15540)

### Definition of Done
- [x] The certificate is not required to be in the files directory

#### Changelog

- [x] This PR does not include user-facing changes and doesn't require a
  CHANGELOG update

#### Test coverage

  changes, or

#### Documentation

#### Behavior

- [x] No behavior was changed with this PR

#### Security


diverdane (approved):
LGTM!!!

PR #443, opened 2022-01-24
ONYX-16318: Add automation for secrets provider p2f authn jwt

tzheleznyak:
Add automation for push2f with authn jwt

Command:
 ./bin/test-workflow/start -n -a secrets-provider-p2f-jwt --jwt

Ticket : ONYX-16318

### DIFF
--- a/helm/conjur-app-deploy/charts/app-secrets-provider-p2f-jwt/templates/NOTES.txt
+++ b/helm/conjur-app-deploy/charts/app-secrets-provider-p2f-jwt/templates/NOTES.txt
@@ -1,0 +1,6 @@
+The Application deployment is complete.
+The following have been deployed:
+- A sample application with a Secrets Provider init container in Push-to-File mode
+  Using authn-jwt flow
+
+Application is now available at test-app-secrets-provider-p2f.{{ .Release.Namespace }}.svc.cluster.local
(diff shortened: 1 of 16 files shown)


helm/conjur-app-deploy/charts/app-secrets-provider-p2f-jwt/templates/NOTES.txt
```diff
@@ -0,0 +1,6 @@
+The Application deployment is complete.
+The following have been deployed:
+- A sample application with a Secrets Provider init container in Push-to-File mode
+- Using authn-jwt flow
```
diverdane, line 4: Minor nit... Can we remove the `-` from this line so it's not a separate list item, i.e.:
```
- A sample application with a Secrets Provider init container in Push-to-File mode
  using authn-jwt flow
```

helm/conjur-app-deploy/charts/app-secrets-provider-p2f-jwt/templates/test_app_secrets_provider_p2f.yaml
```diff
@@ -0,0 +1,130 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: test-app-secrets-provider-p2f
```
diverdane, line 4: Just a question for my own understanding... Not that we should do this now, but at some point in the future, could we consider making it so that both `authn-k8s` and `jwt` app deployments could be done at the same time (even in the same Namespace)? Or is there something fundamental that makes it necessary to keep these deployments mutually exclusive? Obviously, we'd have to enable both `authn-k8s` and `jwt` authenticators in Conjur.

diverdane (commented):
Looks good!!! Minor editorial comment, and a question for my own understanding.

diverdane (approved):
LGTM!!!

PR #445, opened 2022-01-26
ONYX-15085: Allow usage of cluster-prep helm without ServiceAccount a…

tzheleznyak:
This fixes will allow usage of the cluster-prep helm chart in least privilege  way for authn-jwt where cluster role binding and ServiceAccount not needed

For example

```
helm install "cluster-prep" cyberark/conjur-config-cluster-prep  -n "<CONJUR_NAMESPACE>" \ 
     --create-namespace \ 
     --set conjur.account="<CONJUR_ACCOUNT>" \ 
     --set conjur.applianceUrl="<CONJUR_URL>" \ 
     --set conjur.certificateFilePath="<CA FILE PATH>" \  
     --set authnK8s.authenticatorID="<JWT VENDOR NAME>" \
     --set authnK8s.clusterRole.create=false \  
     --set authnK8s.serviceAccount.create=false
```

### DIFF
--- a/helm/conjur-config-cluster-prep/templates/golden_configmap.yaml
+++ b/helm/conjur-config-cluster-prep/templates/golden_configmap.yaml
@@ -20,14 +20,10 @@ data:
     authnK8sAuthenticatorID: {{ required "A valid authnK8s.authenticatorID is required!" .Values.authnK8s.authenticatorID }}
     {{- if eq .Values.authnK8s.clusterRole.create true }}
     authnK8sClusterRole: {{ .Values.authnK8s.clusterRole.name | default "conjur-clusterrole" }}
-    {{- else }}
-    authnK8sClusterRole: {{ required "A valid authnK8s.clusterRole.name is required!" .Values.authnK8s.clusterRole.name }}
     {{- end }}
     authnK8sNamespace: {{ .Release.Namespace }}
     {{- if eq .Values.authnK8s.serviceAccount.create true }}
     authnK8sServiceAccount: {{ .Values.authnK8s.serviceAccount.name | default "conjur-serviceaccount" }}
-    {{- else }}
-    authnK8sServiceAccount: {{ required "A valid authnK8s.serviceAccount.name is required!" .Values.authnK8s.serviceAccount.name }}
     {{- end }}
 
     # Conjur Configuration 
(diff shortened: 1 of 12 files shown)


helm/conjur-config-namespace-prep/README.md
```diff
@@ -69,6 +69,7 @@ The following table lists the configurable parameters of the Conjur Namespace-pr
 |`authnRoleBinding.name`|Name for the RoleBinding generated if the `create` flag is set to `true`.|`"conjur-RoleBinding"`|
 |`authnRoleBinding.create`|Flag to generate the ConfigMap with credentials for accessing Conjur instance.|`true`|
 |`authnRoleBinding.name`|Name for the ConfigMap generated if the `create` flag is set to `true`|`"conjur-configmap"`|
+|`conjurConfigMap.authnStrategy`|Authentication flow to conjur. Can be either `"authn-k8s"` or `"authn-jwt"`|`"authn-k8s"`|
```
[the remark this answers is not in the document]
tzheleznyak: Talked about it

helm/conjur-config-cluster-prep/templates/golden_configmap.yaml
```diff
@@ -20,14 +20,10 @@ data:
     authnK8sAuthenticatorID: {{ required "A valid authnK8s.authenticatorID is required!" .Values.authnK8s.authenticatorID }}
     {{- if eq .Values.authnK8s.clusterRole.create true }}
     authnK8sClusterRole: {{ .Values.authnK8s.clusterRole.name | default "conjur-clusterrole" }}
-    {{- else }}
```
diverdane, line 23: Instead of simply deleting this `{{- else }}` conditional, this should be replaced with an `else if` that looks at the `.Values.conjurConfigMap.authnMethod` setting. If `authnMethod` is not set for JWT, then the `authnK8sClusterRole` setting is still required.

Same for `authnK8sServiceAccount` on line 29 below.

This also means that the 2 UT test cases that were deleted are still needed.

diverdane: Ah, never mind on that ^^^. The `authnMethod` setting doesn't apply to the cluster prep Helm chart.
This is unfortunate, since it opens the door a bit for un-flagged user misconfiguration.

diverdane (approved):
LGTM!

PR #450, opened 2022-02-03
Fix version of bitnami/postgresql helm

tzheleznyak:
Fix version of bitanmi postgresql version so pipelines will pass and won't be broker by any changes of bitnami.

If there is a better solution that will also work in openshift i will close this PR.


diverdane (approved):

szh:
Should we create another issue in Jira to track this so we can revisit when a better fix is available?

PR #451, opened 2022-02-04
ONYX-16318: Move JWT token path to annotation in p2f

tzheleznyak:
In automation of secrets provider with JWT in Push 2 file mod. Move JWT_TOKEN_PATH to annotation 
conjur.org/jwt-token-path


szh:
Why isn't the Jenkins pipeline running on this PR? @ismarc

szh (approved):
LGTM!

PR #455, opened 2022-03-05
Adds poststart hook to delay app startup until SP P2F completes

diverdane:
### Desired Outcome

For the application Helm chart for the Secrets Provider running in Push-to-File mode with secret rotation support, it is desirable that we include:

- A Kubernetes postStart lifecycle hook that delays startup of the application until the SP has finished creating the secret file(s).
- A liveness probe for the application container that forces the application container to be restarted whenever the date of the secrets files change.

### Implemented Changes

The following changes are made to the `app-secrets-provider-rotation` Helm chart in order to delay the application container from starting until after the Secrets Provider sidecar container has finished creating the secret file(s) in the secrets shared volume:

- A 'postStart' lifecycle hook is added to the SP container definition.
- The SP container definition is moved before the app container definition. THe SP container def must appear first in order for the 'postStart' lifecycle hook to take effect before the app container starts.

Also, a liveness probe is added which uses the secret file date to determine if the application should be restarted.

These are being added as a reference for customers who are interested in sequencing the SP sidecar container and their app container startup, and who might need to have their application restarted whenever changes occur in the rendered secret files.

### Connected Issue/Story

Resolves #[relevant GitHub issue(s), e.g. 76]

CyberArk internal issue link: [insert issue ID]()

### Definition of Done
*At least 1 todo must be completed in the sections below for the PR to be
merged.*

#### Changelog

  CHANGELOG update

#### Test coverage

  changes, or

#### Documentation

#### Behavior

#### Security

- [ ] There are no security aspects to these changes

### DIFF
--- a/helm/conjur-app-deploy/charts/app-secrets-provider-rotation/templates/test_app_secrets_provider_rotation.yaml
+++ b/helm/conjur-app-deploy/charts/app-secrets-provider-rotation/templates/test_app_secrets_provider_rotation.yaml
@@ -66,22 +66,6 @@ spec:
     spec:
       serviceAccountName: test-app-secrets-provider-rotation
       containers:
-      - image: {{ printf "%s:%s" .Values.app.image.repository .Values.app.image.tag }}
-        imagePullPolicy: {{ .Values.app.image.pullPolicy }}
-        name: test-app
-        command: [ "java", "-jar", "/app.jar", {{ printf "--spring.config.location=file:%s/application.yaml" .Values.app.secretsMountPath }} ]
-        ports:
-        - name: http
-          containerPort: 8080
-        readinessProbe:
-          httpGet:
-            path: /pets
-            port: http
-          initialDelaySeconds: 15
-          timeoutSeconds: 5
-        volumeMounts:
-        - name: secrets
-          mountPath: {{ .Values.app.secretsMountPath }}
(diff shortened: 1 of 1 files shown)


rpothier (commented):
Looks good! Just waiting for CI to pass.

imheresamir:
@diverdane This is cool. I was thinking... what if SP wrote out a little shell script with the restart logic for the liveness probe somewhere in the mounted volume so that the app container liveness probe just needed to source it to be able to automatically restart when needed? So instead of an inline script it was just `source sp_funcs.sh`. That way if they have their own liveness probe script all they have to do is source that script at the top maybe. May not be a good idea, just a thought I had.

diverdane:
@imheresamir , Hmmm... That's an interesting idea. I was thinking about adding something along those lines, but a little less elegant than what you're proposing. I was thinking about having SP create two sentinel files:

- "secrets-available.txt" : This would only be used by the SP for your `postStart` trick, so it could be written to the same directory as the client tokens get written to (SP's private emptyDir volume)
- "secrets-updated.txt" : This would be written by SP, and monitored by the app container's `livenessProbe`. The existence of the file would trigger a container restart.

But adding script(s) would be make it easier to use!

imheresamir:
A utility script provided by SP could contain the logic for your second point: to check for the existence of the "updated" sentinel. Just trying to think of ways to make it easy for customers to integrate into their existing liveness probe with minimal effort. It could possibly be as simple as a `source sp_utils.sh` one-liner at the top of their script.

PR #466, opened 2022-05-31
Add handling for test image publishing

ismarc:
Intermediate versions of `cyberark/conjur-authn-k8s-cluster-test` need to be published for edge releases, then promoted to latest and proper versions on promotion.  The publish was missing from the edge release section.


neil-k-zero (approved):
Thanks @ismarc, looks good!

PR #480, opened 2022-07-12
Update base images

andytinkham:
Signed-off-by: Andy Tinkham <andy.tinkham@cyberark.com>

### Desired Outcome

Updates test-workflow/test-app-summon/Dockerfile.builder and dev/Dockerfile.debug to later base images.


bin/test-workflow/test_app_summon/Dockerfile.builder
```diff
@@ -1,4 +1,4 @@
-FROM ruby:3.0
+FROM ruby:3.1-slim-bullseye
```
szh, line 1: Is the `slim` image going to be enough for this?

CHANGELOG.md
```diff
@@ -6,6 +6,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+## [0.23.7] - 2022-07-12
+- Updated dev/Dockerfile.debug and test-workflow/test-app-summon/Dockerfile.builder
```
szh, line 10:
```suggestion
### Security
- Updated dev/Dockerfile.debug and test-workflow/test-app-summon/Dockerfile.builder
```

bin/test-workflow/test_app_summon/Dockerfile.builder
szh: Come to think of it, is this even used anywhere?

andytinkham: First step was to build and get obvious failures to see if slim would be enough. But, no, I don't see it used anywhere... Looks like the main Dockerfile makes a builder now instead of using the Dockerfile.builder. I'll try removing it.

szh: Do you think we can just remove the whole 'test_app_summon' folder?

rpothier (approved):
LGTM!

PR #492, opened 2022-12-02
Bump version in changelog so that its higher then the previous promotion

mFelgate:
### Desired Outcome

The latest version in the changelog has to be higher then the latest promotion in jenkins or it will cause the Release builds to fail. This is currently causing the  OCP_NEXT pipeline to fail when testing authn-k8s-client.
### Implemented Changes
Added new version in the changelog file.


gl-johnson:
Wasn't aware of this. In the future, should the PR to add a new changelog version occur prior to promotion? It might be worth adding a note to the documented release process in CONTRIBUTING.md

gl-johnson (approved):

PR #508, opened 2023-04-05
Upgrade ruby from 2.4 to 3.1

juniortaeza:


rpothier (approved):
LGTM!

rpothier (changes requested):
Please update the changelog, otherwise looks good.

rpothier (dismissed):
Thanks for adding this.
It should go in the 0.25.0 section

juniortaeza:
That should fix things, thank you for reviewing!

szh:
@juniortaeza looks like helm tests are failing in CI

john-odonnell:
@szh Helm tests are fixed [here](https://github.com/cyberark/conjur-authn-k8s-client/pull/511), but the build still fails on a CLI bug I'm investigating.

szh:
I just rebased to include the helm fixes from #511

szh:
Hmm, still failing

helm/common/utils.sh
```diff
@@ -93,7 +93,7 @@ function run_helm_unittest() {
     fi
 
     # Run a Helm unit test
-    helm unittest . --helm3
+    helm unittest .
```
john-odonnell, line 96: @szh Because we're pinning the `helm-unittest` version, we need to keep the `--helm3` flag. I'll push a commit updating this.

john-odonnell (approved):

PR #528, opened 2023-11-07
adding proxy support + fixing build script

assafjh:
# Desired Outcome

Conjur API uses the net/http module which supports proxy.
When creating a new HTTPS client we ignore proxy configuration at http.Transport
Implemented Changes

build.sh doesn't work since the script path is not at root folder

# Implemented Changes

http.ProxyFromEnvironment was added to http.Transport - we now support http_proxy, https_proxy, no_proxy from env params.

build.sh script was amended with correct path - script will cd to repo root

### DIFF
--- a/bin/build
+++ b/bin/build
@@ -2,6 +2,9 @@
 
 set -euo pipefail
 
+# go to repo root folder for execution
+cd $(dirname $0)/..
+
 . bin/build_utils
 
 function main() {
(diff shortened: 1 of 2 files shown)


szh:
This is merged in our internal repo and will be included in the next release. Thank you!

szh:
Included in [v0.26.7](https://github.com/cyberark/conjur-authn-k8s-client/releases/tag/v0.26.7)

